PR #266, opened 2022-07-28
allow lag by hour in Join per JoinPart

yunfeng-hao:
1. Allow lag by hour in join per JoinPart. It should be straight forward if we need to change to minutes/seconds.
2. Use rightUnfilledRange instead of the whole leftRange in some cases.
3. I didn't add the lag as param in GroupBy config. We need to verify if the lag is stable across different partitions, otherwise setting a lag per GroupBy could be worse than 1 day lag (best case scenario)

@nikhilsimha @better365 @hzding621


hzding621:
the PR has been open for close to 2 years. `lag` is a good topic for OOC so it's worth revisit at some point. close this PR for inactivity.

PR #294, opened 2022-09-21
[Analyzer] Fix for group_bys with no aggregations

vamseeyarla:
Currently Analyzer fails for the configs that has no aggregations defined.
This fix should take care of it by making sure the GroupBy doesn't raise
a NullPointerException for the group_bys with no aggregations. 

Tested and verified that it works against the configs with no group_bys.


hzding621:
issue has been fixed in main branch. closing this.

PR #339, opened 2022-11-14
Print column name with type conversion errors

nikhilsimha:
We unify types across multiple type systems avro, parquet, thrift etc. During conversion across one to another there are type errors when a type from source system is not supported in the destination system. This PR improves error logging - by printing which column name is causing the conversion issue.


hzding621:
close due to inactivity for 1+ year

PR #378, opened 2023-01-25
Vz  remove count from write partitions

ezvz:
A further simplification to the write partitions logic.


hzding621:
Close due to no update for more than 12 months. We will revisit table write in the future

PR #486, opened 2023-06-06
[ZL][PyApi] Validation for repeated output tables

cristianfr:
## Summary

There are cases where the same group by will be added to a join but with different key mappings.
If mishandled this would lead to the same output tables in separate joinParts.
There's also the case of a user not noticing they are adding the same joinPart twice.

This check should catch that.

## Why / Goal

Prevent users from compiling bad confs.

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @nikhilsimha


hzding621:
close due to inactivity for 1+ yr.

PR #579, opened 2023-10-06
[Fetcher] Print exceptions prior to reporting for metrics

nikhilsimha:
## Summary
We now expose exceptionSampleRate via API - default is 0.1 - one among every 10 exceptions will be logged. 

## Why / Goal
Currently we don't log exceptions at all, which makes debugging difficult. This will make it so that by default we print a fraction of exceptions.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @qiyang0221


pengyu-hou:
Close due to inactivity on it. We can reintroduce the sample rate later on.

PR #633, opened 2023-12-05
Replacing prints with logs

ezvz:
## Summary

Changing all printlns to user a logger.

## Why / Goal
Cleaning up.

## Test Plan
- [x] Local build, local run, unit tests

## Reviewers
@nikhilsimha


pengyu-hou:
looks like this is not needed anymore.

PR #675, opened 2024-01-31
Logging fail open

nikhilsimha:
## Summary
Fail to send log should be a soft failure

## Why / Goal
In staging environments sometimes users won't have access to prod logging topics.

## Test Plan
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update


hzding621:
the current behavior should already be soft failure. exception thrown from within a try block will be captured by try.

Issue #630, opened 2023-12-04
[DISCUSS] CHIP-1 Online IR and GetRequest Caching


caiocamatta-stripe:
Discuss [CHIP-1 Online IR and GetRequest Caching](https://github.com/airbnb/chronon/blob/master/proposals/CHIP-1.md).

Feedback on any aspect of this CHIP is welcome. Let me know if there are any questions!

PR #682, opened 2024-02-12
[CHIP-1] Cache batch IRs in the Fetcher

caiocamatta-stripe:
## Summary

This PR adds a batch IR cache to the Chronon fetcher. 
- In the fetcher, we now cache the decoded IRs for batch GetRequests. When a request comes in, the BaseFetcher will first try to get the IRs from this cache before sending a KV Store request. 
- This cache is shared across all GroupBys being served by the Fetcher.

During our tests, we saw roughly a 4GB increase in memory utilization with a 10,000-element cache, but this number will vary significantly across Chronon users.

This work is part 0, 1, and 2 of [CHIP-1](https://github.com/airbnb/chronon/blob/master/proposals/CHIP-1.md).

#### Enabling the cache

To enable the cache,
1. Chronon users must set the Java arg `ai.chronon.fetcher.batch_ir_cache_size` to a non-zero number. This will create the cache object within Chronon.
2. Enable caching for specific GroupBys via the FlagStore. 

Having a two-step process allows for safer rollouts.

## Why / Goal

Motivation: decrease feature serving latency. 
- At Stripe, we saw up to a 43% decrease in p99 feature serving latency and up to a 30% decrease in CPU utilization after enabling this cache. 

[CHIP-1 – Online IR and GetRequest Caching](https://github.com/airbnb/chronon/blob/master/proposals/CHIP-1.md) & [Discussion](https://github.com/airbnb/chronon/issues/630)

## Test Plan

- [X] Added Unit Tests
- [X] Integration tested (existing FetcherTests)

#### QA Testing
Beyond the code tests, we also tested this change extensively in our QA environment that has a variety of GroupBys. 
- Using some of the additional logging that we have on our version of Chronon, I tested a few different scenarios and confirmed that all the feature values and logs were correct when caching was enabled. 
  - Fetching a key that only has batch data
  - Fetching a key that only has streaming data
  - Fetching a key that has both batch and streaming data
  - Fetching a key that has no data
- Confirmed that no cache gets created or used when the `ai.chronon.fetcher.batch_ir_cache_size` arg is false or not set (status quo).
- Only the GroupBys with `enable_caching` set get cached.

Note that all our GroupBys use [the tiled architecture](https://chronon.ai/Tiled_Architecture.html), but that shouldn't matter here. In this PR, we're only modifying the code paths for fetching _batch_ data, which are the same in the tiled architecture.

#### Online-Offline Consistency
Using two different Joins with 10-15 GroupBys each, we also confirmed that online-offline consistency metrics remained the same before and after enabling caching for all GroupBys.

#### Load tests
We use a long-running large-scale load test (5K requests/sec, 15 GroupBys in the Join) to confirm that these changes are stable. We did not see any error or latency spikes during a multi-day load test with caching enabled.

## Checklist
- [ ] Documentation update

## Reviewers
@nikhilsimha (feel free to assign others instead)
@piyushn-stripe


online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -49,43 +50,57 @@ class FetcherBase(kvStore: KVStore,
...
+    * @param streamingResponsesOpt a response from kv store, if the GroupBy was streaming data.
+    * @param oldServingInfo the GroupByServingInfo used to fetch the GroupBys.
+    * @param queryTimeMs the Request timestamp
+    * @param startTimeMs time when we started fetching the KV store
+    * @param overallLatency the time it took to get the values from the KV store
+    * @param context the Metrics.Context to use for recording metrics
+    * @param totalResponseValueBytes the total size of the response from the KV store
+    * @param keys the keys used to fetch the GroupBy
+    * @return
+    */
```
caiocamatta-stripe, line 71: I'm not sure about leaving these docs in. I might remove them

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,215 @@
...
+            b
+          case null =>
+            println(s"Caching is disabled for $groupByName, enable_caching is not set.")
+            isCachingEnabledForGroupBy.putIfAbsent(groupByName, false)
+            false
+          case _ => false
+        }
+      }
+    )
+  }
```
caiocamatta-stripe, line 60: Once [create feature flags function api #686](https://github.com/airbnb/chronon/pull/686) lands, we can modify `isCachingEnabled` to use feature flags instead. That's much better than a parameter in the GroupBy definition as it allows you to make immediate changes instead of having to modify your GroupBy to enable/disable caching.

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,215 @@
...
+                                                      servingInfo: GroupByServingInfoParsed,
+                                                      keys: Map[String, Any]): Map[String, AnyRef] = {
+    if (!isCachingEnabled(servingInfo.groupBy)) return decodingFunction(batchBytes)
+
+    batchResponses match {
+      case _: KvStoreBatchResponse =>
+        val batchRequestCacheKey =
+          BatchIrCache.Key(servingInfo.groupByOps.batchDataset, keys, servingInfo.batchEndTsMillis)
+        val decodedBytes = decodingFunction(batchBytes)
+        if (decodedBytes != null)
```
caiocamatta-stripe, line 88: Caffeine doesn't allow storing `null`, and in general I decided to exclude negative caching for now. Most likely, the keys that end up cached are not new keys and do have batch data (e.g. big merchants, or power users).

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,215 @@
...
+import scala.util.{Success, Try}
+import java.util.concurrent.ConcurrentHashMap
+import scala.collection.JavaConverters.mapAsScalaConcurrentMapConverter
+import scala.collection.Seq
+
+/*
+ * FetcherCache is an extension to FetcherBase that provides caching functionality. It caches KV store
+ * requests to decrease feature serving latency.
+ * */
+trait FetcherCache {
```
caiocamatta-stripe, line 28: Something I considered months ago when making these code changes was to use a `class` (like "`FetcherBaseCached`") instead of a `trait` . This class would inherit from `FetcherBase` and overrides its methods to add caching. That would arguably be slightly cleaner because it'd give us two completely separate version of the Fetcher. Users would be able to use the normal/old Fetcher with no caching if they wanted. 

This would require a significant amount of additional refactoring and re-testing, and I don't think it's worth it. Ideally, once this IR cache is merged in, it becomes a core part of the fetcher that users can enable / disable for their GroupBys as necessary. We've already tested the status quo (no caching), so IMO the risks that arise from additional refactors would outweigh the benefits of having a separate class.

online/src/main/scala/ai/chronon/online/LRUCache.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import com.github.benmanes.caffeine.cache.{Caffeine, Cache => CaffeineCache}
+
+/**
+  * Utility to create a cache with LRU semantics.
+  *
+  * The original purpose of having an LRU cache in Chronon is to cache KVStore calls and decoded IRs
+  * in the Fetcher. This helps decrease to feature serving latency.
+  */
+object LRUCache {
```
caiocamatta-stripe, line 11: Caffeine is technically not LRU but it's similar. I think naming this `LRUCache` makes it easy to understand what it does (and to contrast it with the existing `TTLCache`).

online/src/main/scala/ai/chronon/online/FetcherBase.scala
piyushn-stripe: I don't mind having these around

online/src/main/scala/ai/chronon/online/FetcherCache.scala
piyushn-stripe: Another challenge with the `FetcherBaseCached` approach is it will lead to a profusion of `FetcherBaseX` and `FetcherBaseY` classes as we keep adding functionalities to the fetcher in the future.

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,215 @@
...
+  // BatchResponses encapsulates either a batch response from kv store or a cached batch response.
+  sealed abstract class BatchResponses {
+    def getBatchBytes(batchEndTsMillis: Long): Array[Byte]
+  }
+  object BatchResponses {
+    def apply(kvStoreResponse: Try[Seq[TimedValue]]): KvStoreBatchResponse = KvStoreBatchResponse(kvStoreResponse)
+    def apply(cachedResponse: FinalBatchIr): CachedFinalIrBatchResponse = CachedFinalIrBatchResponse(cachedResponse)
+    def apply(cachedResponse: Map[String, AnyRef]): CachedMapBatchResponse = CachedMapBatchResponse(cachedResponse)
+  }
+  case class KvStoreBatchResponse(response: Try[Seq[TimedValue]]) extends BatchResponses {
```
piyushn-stripe, line 201: minor - lets add a 1-line comment for all the sealed class based enum options for `BatchResponses`. Useful for a reader to grok the purpose of each of these options without checking the fetcher base code.

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,215 @@
...
+
+  def isCachingEnabled(groupBy: GroupBy): Boolean = {
+    if (!isCacheSizeConfigured || groupBy.getMetaData == null || groupBy.getMetaData.getName == null) return false
+
+    val groupByName = groupBy.getMetaData.getName
+    isCachingEnabledForGroupBy.getOrElse(
+      groupByName, {
+        groupBy.getMetaData.customJsonLookUp("enable_caching") match {
+          case b: Boolean =>
+            println(s"Caching is ${if (b) "enabled" else "disabled"} for $groupByName")
```
piyushn-stripe, line 49: switch to log (here and others?)

piyushn-stripe (approved):
Largely looks good to me as we've reviewed and are running this code internally off our fork. Had a couple of minor cosmetic changes.

online/src/main/scala/ai/chronon/online/FetcherCache.scala
caiocamatta-stripe: Good idea. 

I added comments, let me know if it looks good to you.

piyushn-stripe (approved):

online/src/main/scala/ai/chronon/online/FetcherBase.scala
nikhilsimha: I recommend dropping comments for every param - it adds a lot of noise to code and eventually hurts readability.

I think for a lot of these, the names are descriptive enough. 
In general comments should be either describing very high level control flow, or a specific non obvious behavior.

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -49,43 +50,57 @@ class FetcherBase(kvStore: KVStore,
...
+      case _: CachedBatchResponse => // no-op;
     }
 
-    // bulk upload didn't remove an older batch value - so we manually discard
-    val batchBytes: Array[Byte] = batchResponsesTry
-      .map(_.maxBy(_.millis))
-      .filter(_.millis >= servingInfo.batchEndTsMillis)
-      .map(_.bytes)
-      .getOrElse(null)
+    // The bulk upload may not have removed an older batch values. We manually discard all but the latest one.
```
nikhilsimha, line 93: I like this for example

online/src/main/scala/ai/chronon/online/FetcherBase.scala
caiocamatta-stripe: @nikhilsimha I hear you on the added noise hurting readability. I removed all the `@param`s but left a few comments beside the parameters I think aren't fully descriptive from name alone. Let me know if it looks good to you! :)

online/src/main/scala/ai/chronon/online/FetcherCache.scala
better365: Plus one to Piyush's point here.

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -358,6 +448,9 @@ class FetcherBase(kvStore: KVStore,
       }
   }
 
+  /**
+    * Convert an array of bytes to a FinalBatchIr.
+    */
```
better365, line 453: thank you for adding comments here.

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -196,6 +249,19 @@ class FetcherBase(kvStore: KVStore,
...
   }
 
+  override def isCachingEnabled(groupBy: GroupBy): Boolean = {
+    if (!isCacheSizeConfigured || groupBy.getMetaData == null || groupBy.getMetaData.getName == null) return false
+
+    val isCachingFlagEnabled = flagStore.isSet("enable_fetcher_batch_ir_cache",
+                                               Map("groupby_streaming_dataset" -> groupBy.getMetaData.getName).asJava)
+
+    if (debug)
+      println(
```
better365, line 259:
```suggestion
      logger.info(
```

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -253,15 +319,25 @@ class FetcherBase(kvStore: KVStore,
       }
       request -> groupByRequestMetaTry
     }.toSeq
-    val allRequests: Seq[GetRequest] = groupByRequestToKvRequest.flatMap {
-      case (_, Success(GroupByRequestMeta(_, batchRequest, streamingRequestOpt, _, _))) =>
-        Some(batchRequest) ++ streamingRequestOpt
+
+    // If caching is enabled, we check if any of the GetRequests are already cached. If so, we store them in a Map
+    // and avoid the work of re-fetching them.
```
better365, line 324:
```suggestion
    // If caching is enabled, we check if any of the GetRequests are already cached. If so, we store them in a Map
    // and avoid the work of re-fetching them. It is mainly for batch data requests. 
```

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -253,15 +319,25 @@ class FetcherBase(kvStore: KVStore,
...
+    val cachedRequests: Map[GetRequest, CachedBatchResponse] = getCachedRequests(groupByRequestToKvRequest)
+    // Collect cache metrics once per fetchGroupBys call; Caffeine metrics aren't tagged by groupBy
+    maybeBatchIrCache.foreach(cache =>
+      LRUCache.collectCaffeineCacheMetrics(caffeineMetricsContext, cache.cache, cache.cacheName))
+
+    val allRequestsToFetch: Seq[GetRequest] = groupByRequestToKvRequest.flatMap {
+      case (_, Success(GroupByRequestMeta(_, batchRequest, streamingRequestOpt, _, _))) => {
+        // If a request is cached, don't include it in the list of requests to fetch.
```
better365, line 332:
```suggestion
        // If a batch request is cached, don't include it in the list of requests to fetch because the batch IRs already cached
```

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,209 @@
...
+/*
+ * FetcherCache is an extension to FetcherBase that provides caching functionality. It caches KV store
+ * requests to decrease feature serving latency.
+ * */
+trait FetcherCache {
+  @transient private lazy val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  val batchIrCacheName = "batch_cache"
+  val maybeBatchIrCache: Option[BatchIrCache] =
+    Option(System.getProperty("ai.chronon.fetcher.batch_ir_cache_size"))
```
better365, line 34: what is the unit for the size? Mb or elements size?

online/src/test/scala/ai/chronon/online/FetcherCacheTest.scala
```diff
@@ -0,0 +1,334 @@
...
+}
+
+class FetcherCacheTest extends MockitoHelper {
+  class TestableFetcherCache(cache: Option[BatchIrCache]) extends FetcherCache {
+    override val maybeBatchIrCache: Option[BatchIrCache] = cache
+  }
+  val batchIrCacheMaximumSize = 50
+
+  @Test
+  def test_BatchIrCache_CorrectlyCachesBatchIrs(): Unit = {
```
better365, line 37: sorry for being pedant, the code base uses camel cases only for the naming convention. Could you rename them to be camel case? Thanks!

online/src/test/scala/ai/chronon/online/FetcherBaseTest.scala
```diff
@@ -141,4 +148,76 @@ class FetcherBaseTest extends MockitoSugar with Matchers {
     actualRequest.get.name shouldBe query.groupByName + "." + query.columnName
     actualRequest.get.keys shouldBe query.keyMapping.get
   }
+
+  // updateServingInfo() is called when the batch response is from the KV store.
+  @Test
+  def test_getServingInfo_ShouldCallUpdateServingInfoIfBatchResponseIsFromKvStore(): Unit = {
```
better365, line 154: Same as the other one, could you make it camel case only to be consistent with the naming convention? thank you!

online/src/main/scala/ai/chronon/online/LRUCache.scala
```diff
@@ -0,0 +1,57 @@
...
+    * @return Caffeine cache
+    */
+  def apply[KEY <: Object, VALUE <: Object](cacheName: String, maximumSize: Int = 10000): CaffeineCache[KEY, VALUE] = {
+    buildCaffeineCache[KEY, VALUE](cacheName, maximumSize)
+  }
+
+  private def buildCaffeineCache[KEY <: Object, VALUE <: Object](
+      cacheName: String,
+      maximumSize: Int = 10000): CaffeineCache[KEY, VALUE] = {
+    println(s"Chronon Cache build started. cacheName=$cacheName")
```
better365, line 29:
```suggestion
    logger.info(s"Chronon Cache build started. cacheName=$cacheName")
```

online/src/main/scala/ai/chronon/online/LRUCache.scala
```diff
@@ -0,0 +1,57 @@
...
+    val cache: CaffeineCache[KEY, VALUE] = Caffeine
+      .newBuilder()
+      .maximumSize(maximumSize)
+      .recordStats()
+      .build[KEY, VALUE]()
+    println(s"Chronon Cache build finished. cacheName=$cacheName")
```
better365, line 35:
```suggestion
    logger.info(s"Chronon Cache build finished. cacheName=$cacheName")
```

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -196,6 +249,19 @@ class FetcherBase(kvStore: KVStore,
     }
   }
 
+  override def isCachingEnabled(groupBy: GroupBy): Boolean = {
+    if (!isCacheSizeConfigured || groupBy.getMetaData == null || groupBy.getMetaData.getName == null) return false
+
+    val isCachingFlagEnabled = flagStore.isSet("enable_fetcher_batch_ir_cache",
```
better365, line 255: could the flagstore be null here as the default value? 
Could you add more details about how the flag store information is being created? 

If the flags are getting more and more, we should consider moving all the flags with its functionality to a single file.

better365:
should Add enable_caching=True to their GroupBy definition actually be 

`Add enable_fetcher_batch_ir_cache=True to their GroupBy definition`?

One gap here is how the flagStore is created. I assume it is from metadata?

online/src/main/scala/ai/chronon/online/FetcherBase.scala
caiocamatta-stripe: Good point. I think it could be `null`. 

The FlagStore is passed in via the Chronon API. Users can override with their own implementation. https://github.com/airbnb/chronon/pull/686. I'll add some more information in the flagstore file to clarify.

caiocamatta-stripe:
Hey @pengyu-hou, I addresses your comments. Could you take another look?


> should Add enable_caching=True to their GroupBy definition actually be Add enable_fetcher_batch_ir_cache=True to their GroupBy definition?

I updated the PR description. This `Add enable_caching=True to their GroupBy definition` comment was related to the previous version of this PR which used `customJson` in the GroupBy instead of the FlagStore.

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -196,6 +249,20 @@ class FetcherBase(kvStore: KVStore,
     }
   }
 
+  override def isCachingEnabled(groupBy: GroupBy): Boolean = {
+    if (!isCacheSizeConfigured || groupBy.getMetaData == null || groupBy.getMetaData.getName == null) return false
+
+    val isCachingFlagEnabled =
+      flagStore != null && flagStore.isSet("enable_fetcher_batch_ir_cache",
+                                           Map("groupby_streaming_dataset" -> groupBy.getMetaData.getName).asJava)
```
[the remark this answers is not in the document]
pengyu-hou: In a follow up PR (either from us or stripe), we could consider moving the all flags to a dedicated file like Constants.

online/src/main/scala/ai/chronon/online/FetcherCache.scala
```diff
@@ -0,0 +1,216 @@
...
+ * small number (e.g. 1,000) and measure how much memory it uses, then adjust accordingly.
+ *  2. Enable caching for a specific GroupBy by overriding `isCachingEnabled` and returning `true` for that GroupBy.
+ * FetcherBase already provides an implementation of `isCachingEnabled` that uses the FlagStore.
+ * */
+trait FetcherCache {
+  @transient private lazy val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  val batchIrCacheName = "batch_cache"
+  val maybeBatchIrCache: Option[BatchIrCache] =
+    Option(System.getProperty("ai.chronon.fetcher.batch_ir_cache_size_elements"))
```
pengyu-hou, line 41: We may need to refactor this when we start integration

pengyu-hou (approved):
Thanks @caiocamatta-stripe for contributing back to Chronon repo! 

Please address the last comment before merging, thanks!

PR #686, opened 2024-02-21
flagstore interface

divyamanohar-stripe:
## Summary

Create a function in Api.scala that users can override to enable feature flag functionality

## Why / Goal

Ability to safely roll out features in a controlled manner

## Test Plan

- [x] Covered by existing CI
- [x] tested in our environment, passing in our feature flag functionality to our Api class and verified we can turn off/on features in Chronon code


online/src/main/scala/ai/chronon/online/Api.scala
```diff
@@ -190,20 +190,34 @@ abstract class Api(userConf: Map[String, String]) extends Serializable {
...
+                    responseConsumer,
+                    externalRegistry,
+                    featureFlagBiPredicate)
 
   private def responseConsumer: Consumer[LoggableResponse] =
     new Consumer[LoggableResponse] {
       override def accept(t: LoggableResponse): Unit = logResponse(t)
     }
+
+  private def featureFlagBiPredicate: BiPredicate[String, java.util.Map[String, String]] =
```
nikhilsimha, line 219: I wonder if having a class called flagStore, the right interface. It might be easier to evolve it in the future.

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -81,7 +80,8 @@ class Fetcher(val kvStore: KVStore,
               timeoutMillis: Long = 10000,
               logFunc: Consumer[LoggableResponse] = null,
               debug: Boolean = false,
-              val externalSourceRegistry: ExternalSourceRegistry = null)
+              val externalSourceRegistry: ExternalSourceRegistry = null,
```
nikhilsimha, line 83: This is not backwards compatible IMO. I think we should make this nullable, and use get set methods to modify the flagStore

divyamanohar-stripe:
@nikhilsimha I've made the featureFlags param in `Fetcher.scala` nullable and it's now backwards compatible. I'm following the general same pattern of `responseConsumer` in `Api.scala` for this, lmk what you think

nikhilsimha (approved):
Looks clean

caiocamatta-stripe (approved):
Looks great!

PR #716, opened 2024-03-12
Derivation error handling to be non-blocking and return partial responses 

better365:
## Summary

For derivation failures, the current behavior will throw an exception. It will block the program for online feature serving. Therefore, we want to make it non-blocking for online serving path.

This PR will turn an exception map rather than throwing out. And, use Try to handle more errors even during join codec build stage to guarantee non-blocking for online env. 

Minor change: refactor the logger to be in the metrics.scala so that error messages will be logged as well. 

## Why / Goal

Make online path more stable. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update


online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -190,23 +190,40 @@ class Fetcher(val kvStore: KVStore,
...
+                ctx.incrementException(joinTry.failed.get)
+                (Map.empty[String, AnyRef], Map.empty[String, AnyRef])
+              } else {
+                val join = joinTry.get
+                if (join.derivationsContainStar && !join.areDerivationsRenameOnly) {
+                  // if it is not rename only with a star in it, we need to return the base map
+                  (Map.empty[String, AnyRef], baseMap)
+                } else {
+                  (derivedExceptionMap, Map.empty[String, AnyRef])
+                }
```
better365, line 220: @nikhilsimha  This is the code block when the derivation try is failed. Can you take a look?

better365: cc: @hzding621 can you also take a look here? Thanks!

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -190,23 +190,40 @@ class Fetcher(val kvStore: KVStore,
...
                 .deriveFunc(internalResponse.request.keys, baseMap ++ tsDsMap)
                 .mapValues(_.asInstanceOf[AnyRef])
-                .toMap) match {
-              case Success(derivedMap) => derivedMap
-              case Failure(exception) => {
-                ctx.incrementException(exception)
-                throw exception
+            }
+
+            val (derivedMapCleaned, baseMapCleaned) = if (derivedMapTry.isSuccess) {
```
nikhilsimha, line 201: use case match instead of if else.

better365: will do

better365: Done

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -190,23 +190,38 @@ class Fetcher(val kvStore: KVStore,
             val baseMap = internalMap ++ externalMap
             // used for derivation based on ts/ds
             val tsDsMap: Map[String, AnyRef] =
-              Map("ts" -> (requestTs).asInstanceOf[AnyRef], "ds" -> (requestDs).asInstanceOf[AnyRef])
-            val derivedMap: Map[String, AnyRef] = Try(
+              Map("ts" -> requestTs.asInstanceOf[AnyRef], "ds" -> requestDs.asInstanceOf[AnyRef])
```
hzding621, line 193: Given this change, `ts` and `ds` are now reserved keywords. Can we add some validations in join.py when creating `Derivation` to prevent people from using `ts` and `ds` as names of derived features? Otherwise this will introduce bugs that are hard to debug.

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -190,23 +190,38 @@ class Fetcher(val kvStore: KVStore,
...
+                getJoinConf(joinName) match {
+                  case Failure(joinException) =>
+                    ctx.incrementException(joinException)
+                    (Map.empty[String, AnyRef], Map.empty[String, AnyRef])
+                  case Success(join) =>
+                    if (join.derivationsContainStar) {
+                      (derivedExceptionMap, baseMap)
+                    } else {
+                      (derivedExceptionMap, Map.empty[String, AnyRef])
+                    }
```
hzding621, line 218: this should just be 
```
(derivedExceptionMap, baseMap)
```

we want to keep `baseMap` even if derivations does not contain star.

online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: Not have to be in this PR

better365: Sounds good. Let me add that!

online/src/main/scala/ai/chronon/online/Fetcher.scala
better365: @nikhilsimha I feel like the same way Nikhil. Should we simply return baseMap always `(derivedExceptionMap, baseMap)`?

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -190,23 +190,38 @@ class Fetcher(val kvStore: KVStore,
...
+              case Success(derivedMap) =>
+                val cleanedDerivedMap = derivedMap -- tsDsMap.keys
+                (cleanedDerivedMap, baseMap)
+              case Failure(exception) =>
                 ctx.incrementException(exception)
-                throw exception
-              }
+                val derivedExceptionMap =
+                  Map("derivation_fetch_exception" -> exception.traceString.asInstanceOf[AnyRef])
+                getJoinConf(joinName) match {
```
hanyuli1995, line 209: Just curious, what is the purpose of calling getJoinConf here? In which case this function will return failure?

nikhilsimha: we would have gotten the joinConf once already right? We should just use that. 

This could have updated meanwhile if the joinCodec has changed since the first fetch. Also that will make the code simpler to deal with.

better365: Got it. Because the fetch results from internal and external don't have the join config itself. 

but if we simply return baseMap anyways, we don't need to check if there is a star in derivations or not.

online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: If you check the downstream usage of `derivedMapCleaned` and `baseMapCleaned`: `
- `derivedMapCleaned` is returned back to user in responseMap, and thus is consistent to what users want to see
- `baseMapCleaned` is only used during logging.

hzding621 (approved):

online/src/main/scala/ai/chronon/online/Fetcher.scala
better365: Thanks Haozhen. I renamed the variables and cleaned it up.

nikhilsimha (approved):
Looks clean!

online/src/main/scala/ai/chronon/online/Fetcher.scala
better365: Per the above comment, the join conf is not needed here. Getting the conf again can be deleted.

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -183,30 +193,46 @@ class Fetcher(val kvStore: KVStore,
...
+                val requestDs = TsUtils.toStr(requestTs).substring(0, 10)
+                val baseMap = internalMap ++ externalMap
+                // used for derivation based on ts/ds
+                val tsDsMap: Map[String, AnyRef] =
+                  Map(Constants.TimeColumn -> requestTs, "ds" -> requestDs)
+                    .mapValues(_.asInstanceOf[AnyRef])
+                    .toMap
+                val derivedMapTry: Try[Map[String, AnyRef]] = Try {
+                  joinCodec
+                    .deriveFunc(internalResponse.request.keys, baseMap ++ tsDsMap)
```
hanyuli1995, line 210: Based on the code, if the user defined derivation is incorrect(which will fail the derievFunc initialization), this step will fail and we will still send out the partial result with exception message right?

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -183,30 +193,46 @@ class Fetcher(val kvStore: KVStore,
...
+
+                val derivedMap = derivedMapTry match {
+                  case Success(derivedMap) =>
+                    val derivedCleanedMap = derivedMap -- tsDsMap.keys
+                    derivedCleanedMap
+                  case Failure(exception) =>
+                    ctx.incrementException(exception)
+                    val derivedExceptionMap =
+                      Map("derivation_fetch_exception" -> exception.traceString.asInstanceOf[AnyRef])
+                    derivedExceptionMap
```
hanyuli1995, line 223: @better365  Question, should we include the baseMap here? My understanding is this derivedMap defined in ResponseWithContext is the complete result after derivation, it includes both the original value(except the columns which are renamed) and the derived new columns. So even if the deriveFunc failed we still want to keep the baseMap here along with the exception message right? Pls let me know if my understanding is incorrect here.

hanyuli1995 (approved):

online/src/main/scala/ai/chronon/online/Fetcher.scala
better365: After discussion with the team, we will apply rename only derivations at this step. If client has a fall back logic, it will make no difference. Otherwise, it will be good to have partial responses.

PR #724, opened 2024-03-16
[Fetcher] Don't throw for old query time; Use datadog to count it instead

better365:
## Summary

We got false alarm caused by old query time. In this PR, we will change it to use datadog to count the exceptions. 

## Why / Goal

Avoid false alarms. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hanyuli1995 @donghanz

### DIFF
--- a/online/src/main/scala/ai/chronon/online/FetcherBase.scala
+++ b/online/src/main/scala/ai/chronon/online/FetcherBase.scala
@@ -82,7 +82,12 @@ class FetcherBase(kvStore: KVStore,
       val streamingResponses = streamingResponsesOpt.get
       val mutations: Boolean = servingInfo.groupByOps.dataModel == DataModel.Entities
       val aggregator: SawtoothOnlineAggregator = servingInfo.aggregator
-      if (batchBytes == null && (streamingResponses == null || streamingResponses.isEmpty)) {
+      if (aggregator.batchEndTs > queryTimeMs) {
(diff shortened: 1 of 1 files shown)


hzding621 (approved):

aggregator/src/main/scala/ai/chronon/aggregator/windowing/SawtoothOnlineAggregator.scala
```diff
@@ -89,10 +89,6 @@ class SawtoothOnlineAggregator(val batchEndTs: Long,
     val batchIr = Option(finalBatchIr).getOrElse(normalizeBatchIr(init))
     val headRows = Option(streamingRows).getOrElse(Array.empty[Row].iterator)
 
-    if (batchEndTs > queryTs) {
```
nikhilsimha, line 92: If we get here, we will instead throw an NPE right? Let's keep this around. Your other changes will gate the control from throwing this anyways.

better365: sounds good.

hanyuli1995 (approved):

donghanz (approved):

PR #735, opened 2024-03-29
Add APPROX_HISTOGRAM_K Operation

jbrooks-stripe:
## Summary

Adds an APPROX_HISTOGRAM_K operation based on the FrequentItems Sketch:
https://datasketches.apache.org/docs/Frequency/FrequentItemsOverview.html

This is mostly a wrapper on top of the frequent items sketch, but have made the two following modifications to make it somewhat more user friendly:
* The sketch requires a k which is a power of two. Instead of making this a user requirement, we will take the next closest power of two and truncate down to k in the results.
* The underlying map used by the sketch has a load factor of 0.75, which causes the approximation to kick in before there are k elements. We changed this to keep an exact histogram/map and then switch over to the sketch once the entries exceed k.

## Why / Goal

The current histogram operation uses unbounded memory and isn't stable for production use cases. 

## Test Plan

We've end to end tested everything through backfills/CLI on the batch side on our end.

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [x] Documentation update


aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -1,35 +1,14 @@
-/*
```
nikhilsimha, line 1: Is this intentional?

aggregator/src/main/scala/ai/chronon/aggregator/row/ColumnAggregator.scala
```diff
@@ -375,6 +377,17 @@ object ColumnAggregator {
       case Operation.LAST    => timed(new Last(inputType))
       case Operation.FIRST_K => timed(new FirstK(inputType, aggregationPart.getInt("k")))
       case Operation.LAST_K  => timed(new LastK(inputType, aggregationPart.getInt("k")))
+      case Operation.APPROX_HISTOGRAM_K =>
```
nikhilsimha, line 380: Can you move this up closer to histogram?

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -459,8 +465,7 @@ class ApproxDistinctCount[Input: CpcFriendly](lgK: Int = 8) extends SimpleAggreg
     CpcSketch.heapify(normalized.asInstanceOf[Array[Byte]])
 }
 
-class ApproxPercentiles(k: Int = 128, percentiles: Array[Double] = Array(0.5))
-    extends SimpleAggregator[Float, KllFloatsSketch, Array[Float]] {
+class ApproxPercentiles(k: Int = 128, percentiles: Array[Double] = Array(0.5)) extends SimpleAggregator[Float, KllFloatsSketch, Array[Float]] {
```
nikhilsimha, line 468: Intentional?

spark/src/main/scala/ai/chronon/spark/ChrononKryoRegistrator.scala
```diff
@@ -1,28 +1,15 @@
-/*
```
nikhilsimha, line 1: Intentional?

nikhilsimha (commented):
Few comments. But the approach looks sound overall.

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
jbrooks-stripe: Not intentional. Updating merge conflicts. Will tag you when everything is ready.

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -372,7 +373,12 @@ class FrequentItems[T: FrequentItemsFriendly](val mapSize: Int, val errorType: E
   override def irType: DataType = BinaryType
   type Sketch = ItemsSketch[T]
   override def prepare(input: T): Sketch = {
-    val sketch = new ItemsSketch[T](mapSize)
+    // The ItemsSketch implementation requires a size with a positive power of 2
+    // Initialize the sketch with the next closest power of 2
+    val power = math.ceil(math.log(math.max(mapSize, 1)) / math.log(2))
```
nikhilsimha, line 378: ChatGpt gave me 

int nextPowerOf2 = (n > 1) ? Integer.highestOneBit(n - 1) << 1 : 1;

Seems more readable I think.

spark/src/main/scala/ai/chronon/spark/ChrononKryoRegistrator.scala
```diff
@@ -127,7 +126,6 @@ class ChrononKryoRegistrator extends KryoRegistrator {
       "scala.reflect.ManifestFactory$LongManifest",
       "org.apache.spark.sql.execution.joins.EmptyHashedRelation$",
       "scala.reflect.ManifestFactory$$anon$1",
-      "scala.reflect.ClassTag$GenericClassTag",
```
nikhilsimha, line 130: This might cause issues based on version of scala we are testing with. We recently disabled scala 2.11 tests - which is probably the version that removing this will break.

Let's remove this line.

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
nikhilsimha, line 378: It might be worth adding tests to this line at various boundary conditions - -1, 0, 1, 31, 32, 33

nikhilsimha (commented):
Let's also add a test to spark/test/FetcherTest.scala 

Looks good otherwise. Mostly non-blocking comments.

jbrooks-stripe:
> Let's also add a test to spark/test/FetcherTest.scala
> 
> Looks good otherwise. Mostly non-blocking comments.

Appreciate you taking a look. Will address your comments and some from our internal repo shortly.

andrewlee-stripe (approved):

PR #739, opened 2024-04-05
Fix bug in CatalystUtil causing `where`s to not be applied

caiocamatta-stripe:
## Summary

Fixes an issue in CatalystUtil that causes the `where` clauses in some GroupBys to not be applied.

Some notes from @piyushn-stripe's investigation: When certain `select`s exist in a GroupBy (such as the `CAST(get_json_object(...` in the unit test we added to this PR), we were hitting the `ProjectExec(projectList, childPlan)` branch of CatalystUtil `initialize` code instead of `case ProjectExec(projectList, fp@FilterExec(condition, child))`. This caused the `where`s to not be applied.

To fix this, we added code from the `WholeStageCodecgenExec` branch into the `ProjectExec(projectList, childPlan)` branch.

## Why / Goal

We observed that our Flink app ([Chronon on Flink](https://chronon.ai/setup/Flink.html)) was not filtering out events and the `where` clause in some GroupBys wasn't being applied at all. We narrowed it down to an issue in CatalystUtil and were able to reproduce it with a unit test.

## Test Plan

- [X] Added Unit Tests
- [X] Covered by existing CI (to some extent; there are a number of existing CatalystUtil tests)
- [ ] Integration tested

We temporarily modified the code here to throw an error in the new `case whc @ WholeStageCodegenExec(fp @ FilterExec(condition, child)) =>` case, and verified that only the GroupBys that weren't filtering correctly are affected by this change. We have hundreds of GroupBys defined, so this gives us confidence that, at least on our side, this is a safe change.

## Checklist
- [ ] Documentation update

### DIFF
--- a/online/src/main/scala/ai/chronon/online/CatalystUtil.scala
+++ b/online/src/main/scala/ai/chronon/online/CatalystUtil.scala
@@ -161,7 +161,7 @@ class CatalystUtil(expressions: collection.Seq[(String, String)],
 
     // extract transform function from the df spark plan
     val func: InternalRow => Option[InternalRow] = filteredDf.queryExecution.executedPlan match {
-      case whc: WholeStageCodegenExec =>
+      case whc: WholeStageCodegenExec => {
         val (ctx, cleanedSource) = whc.doCodeGen()
         val (clazz, _) = CodeGenerator.compile(cleanedSource)
         val references = ctx.references.toArray
@@ -176,6 +176,7 @@ class CatalystUtil(expressions: collection.Seq[(String, String)],
           None
         }
         codegenFunc
+      }
       case ProjectExec(projectList, fp @ FilterExec(condition, child)) => {
         val unsafeProjection = UnsafeProjection.create(projectList, fp.output)
 
@@ -191,6 +192,7 @@ class CatalystUtil(expressions: collection.Seq[(String, String)],
       }
       case ProjectExec(projectList, childPlan) => {
         childPlan match {
+          // This WholeStageCodegenExec case is slightly different from the one above as we apply a projection.
           case whc @ WholeStageCodegenExec(fp: FilterExec) =>
             val unsafeProjection = UnsafeProjection.create(projectList, childPlan.output)
             val (ctx, cleanedSource) = whc.doCodeGen()


online/src/main/scala/ai/chronon/online/CatalystUtil.scala
```diff
@@ -191,11 +191,31 @@ class CatalystUtil(expressions: collection.Seq[(String, String)],
         projectFunc
       }
       case ProjectExec(projectList, childPlan) => {
-        val unsafeProjection = UnsafeProjection.create(projectList, childPlan.output)
-        def projectFunc(row: InternalRow): Option[InternalRow] = {
-          Some(unsafeProjection.apply(row))
+        childPlan match {
+          case whc @ WholeStageCodegenExec(fp @ FilterExec(condition, child)) =>
```
caiocamatta-stripe, line 195: I can remove code repetition with the other `WholeStageCodegenExec` branch, but the general idea is here & folks can review.

online/src/test/scala/ai/chronon/online/test/CatalystUtilTest.scala
```diff
@@ -584,4 +584,30 @@ class CatalystUtilTest extends TestCase with CatalystUtilTestSparkSQLStructs {
...
+    )
+    val selects = Map(
+      "id" -> "key",
+      "created" -> "created_ts",
+      "score" -> "CAST(get_json_object(json_prediction, '$.score') as Double)"
+    ).toSeq
+    val wheres = Seq("tag = 'inexistent'")
+    val cu = new CatalystUtil(selects, inputEventStruct, wheres)
+    val res = cu.performSql(inputEventRow)
+    assertTrue(res.isEmpty)
```
caiocamatta-stripe, line 611: Without the CatalystUtil changes, this fails.

nikhilsimha (approved):
Thanks for the fix and for the recent commit to reduce duplication.

How did the bug go in? -  only flink code uses catalyst util for filtering. I thought that tests filtering.

online/src/test/scala/ai/chronon/online/test/CatalystUtilTest.scala
```diff
@@ -584,4 +584,30 @@ class CatalystUtilTest extends TestCase with CatalystUtilTestSparkSQLStructs {
     assertTrue(res.get("c").asInstanceOf[util.ArrayList[Any]].contains("world"))
   }
 
+  def testWhereClauseShouldFilterEventOut(): Unit = {
```
piyushn-stripe, line 587: @caiocamatta-stripe - should we add a test that covers the case where we include the row and confirm the projected columns etc match what we expect?

piyushn-stripe (approved):

online/src/main/scala/ai/chronon/online/CatalystUtil.scala
```diff
@@ -191,11 +190,30 @@ class CatalystUtil(expressions: collection.Seq[(String, String)],
         projectFunc
       }
       case ProjectExec(projectList, childPlan) => {
-        val unsafeProjection = UnsafeProjection.create(projectList, childPlan.output)
-        def projectFunc(row: InternalRow): Option[InternalRow] = {
-          Some(unsafeProjection.apply(row))
+        childPlan match {
```
piyushn-stripe, line 193: should we add a comment here indicating that this is subtly different from the prior codegen block?

piyushn-stripe (approved):

PR #740, opened 2024-04-09
Add skew and kurtosis operations

jbrooks-stripe:
## Summary

Adds operations for [skew](https://en.wikipedia.org/wiki/Skew) and [kurtosis](https://en.wikipedia.org/wiki/Kurtosis). 

The implementation uses the same Welford/Knuth approach as the variance operation.
https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Higher-order_statistics

I have the operations set as non-deletable as the reversals weren't straightforward (variance is also set this way). 

## Why / Goal

We have existing features in our previous feature store which rely on these statistics. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [x] Documentation update

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
@@ -770,8 +770,6 @@ case class MomentsIR(
 )
 
 // Uses Welford/Knuth method as the traditional sum of squares based formula has serious numerical stability problems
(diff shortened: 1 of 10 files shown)


nikhilsimha (approved):

nikhilsimha:
Solid PR!

andrewlee-stripe (approved):

piyushn-stripe (approved):

PR #743, opened 2024-04-11
Unsetting wheres for left query when using bootstrap and adding joinSource logic

ezvz:
## Summary
Unsetting `wheres` when using bootstrap because wheres already applied to left side.
Implementing table overwrite for bootstrap for joinSource case.

## Why / Goal
Fixes column not found bug.

## Test Plan

- [ ] Devbox test

## Reviewers
@hzding621 @donghanz

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
+++ b/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
@@ -411,7 +411,6 @@ abstract class JoinBase(joinConf: api.Join,
     }
 
     val source = joinConf.left
-
     if (useBootstrapForLeft) {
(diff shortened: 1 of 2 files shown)


nikhilsimha (approved):

spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -411,15 +411,14 @@ abstract class JoinBase(joinConf: api.Join,
     }
 
     val source = joinConf.left
-    if (useBootstrapForLeft && !source.isSetJoinSource) {
+
```
ezvz, line 414:
```suggestion
```

PR #752, opened 2024-04-24
Update analyzer for metadata extraction

hanyuli1995:
## Summary

The metadata exporter is calling Analyzer to do data analysis before generating the metadata, which include:
1. Validate input source table permission for the users who run the metadata export.
2. Run a query on the input source table to get the output schema.

The process works well for users, but for metadata exporter since it is running by chronon team who may not have access on some of the source tables, the task will fail to generate the metadata for these group_by/joins. 

To solve this issue, we decided to introduce a flag **validateTablePermission** in Analyzer for metadata exporter. The flag is set to be true by default. But it is set to be false for metadata exporter and it will avoid table permission validation and source table read to get rid of the permission issue. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@ezvz @nikhilsimha

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -257,10 +257,20 @@ class Analyzer(tableUtils: TableUtils,
     // run SQL environment setups such as UDFs and JARs
     joinConf.setups.foreach(tableUtils.sql)
 
-    val leftDf = JoinUtils.leftDf(joinConf, range, tableUtils, allowEmpty = true).get
-    val analysis = if (enableHitter) analyze(leftDf, joinConf.leftKeyCols, joinConf.left.table) else ""
-    val leftSchema: Map[String, DataType] =
-      leftDf.schema.fields.map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType))).toMap
+    val (analysis, leftSchema) = if (enableHitter) {
+      val leftDf = JoinUtils.leftDf(joinConf, range, tableUtils, allowEmpty = true).get
+      val analysis = analyze(leftDf, joinConf.leftKeyCols, joinConf.left.table)
+      val leftSchema: Map[String, DataType] =
+        leftDf.schema.fields.map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType))).toMap
+      (analysis, leftSchema)
+    } else {
+      val analysis = ""
+      val scanQuery = range.genScanQuery(joinConf.left.query,
+        joinConf.left.table,
+        fillIfAbsent = Map(tableUtils.partitionColumn -> null))
+      val leftSchema: Map[String, DataType] = tableUtils.sql(scanQuery).schema.fields.map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType))).toMap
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -251,16 +251,34 @@ class Analyzer(tableUtils: TableUtils,
...
-    val analysis = if (enableHitter) analyze(leftDf, joinConf.leftKeyCols, joinConf.left.table) else ""
-    val leftSchema: Map[String, DataType] =
-      leftDf.schema.fields.map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType))).toMap
+    val (analysis, leftSchema) = if (enableHitter) {
+      val leftDf = JoinUtils.leftDf(joinConf, range, tableUtils, allowEmpty = true).get
+      val analysis = analyze(leftDf, joinConf.leftKeyCols, joinConf.left.table)
+      val leftSchema: Map[String, DataType] =
+        leftDf.schema.fields.map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType))).toMap
+      (analysis, leftSchema)
+    } else if (metadataExtraction){
```
ezvz, line 268: I think we can simplify this logic. Basically we only do the analysis for heavy hitters. Otherwise I think we can use the same schema extraction logic for all flows.

ezvz (commented):
Let's make the flag `validateTablePermission` to make it more explicit. And given my understanding that query.schema shouldn't trigger data access issues, I don't think we need the 3x if/else statement.

ezvz:
@hanyuli1995 small nit, but let's avoid internal slack links in the open source channel. It would be good to have context here that anyone can read and understand.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -251,17 +251,34 @@ class Analyzer(tableUtils: TableUtils,
...
     logger.info(s"""|Running join analysis for $name ...""".stripMargin)
     // run SQL environment setups such as UDFs and JARs
     joinConf.setups.foreach(tableUtils.sql)
 
-    val leftDf = JoinUtils.leftDf(joinConf, range, tableUtils, allowEmpty = true).get
-    val analysis = if (enableHitter) analyze(leftDf, joinConf.leftKeyCols, joinConf.left.table) else ""
-    val leftSchema: Map[String, DataType] =
-      leftDf.schema.fields.map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType))).toMap
-
+    val (analysis, leftSchema) = if (enableHitter) {
```
ezvz, line 261: One more suggestion to simplify, you can return `leftDf` here instead of `leftSchema`. Then you don't need to duplicate this logic:

```
        .schema
        .fields
        .map(field => (field.name, SparkConversions.toChrononType(field.name, field.dataType)))
        .toMap
```

SophieYu41 (approved):
LGTM. Thanks for the change!

I would also test regular user Analyzer flow making sure the permission flag is "True"

ezvz (approved):

PR #753, opened 2024-04-24
FetcherCLI: support fetching with local join conf file without uploading to metadata store

better365:
## Summary

This PR will enable users to use a local join conf file to fetch features without uploading the conf to metadata store.

## Why / Goal

It will simplify the testing process. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [x] tested on gateway macine
```
python3 ./scripts/run.py --mode=fetch -k '<place_holder>' --conf-path <place_holder> -t join --chronon-jar ~/test/chronon-embedded.jar

2024-04-24 21:09:16 INFO  Fetcher:391 - Using passed in join configuration: <place_holder>
```

## Checklist
- [ ] Documentation update

## Reviewers
@airbnb/zipline-maintainers @nikhilsimha @hanyuli1995 @donghanz

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Driver.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Driver.scala
@@ -594,8 +594,9 @@ object Driver {
     @transient lazy val logger = LoggerFactory.getLogger(getClass)
 
     class Args extends Subcommand("fetch") with OnlineSubcommand {
+      val confPath: ScallopOption[String] = opt[String](required = false, descr = "Path to conf to fetch features")
       val keyJson: ScallopOption[String] = opt[String](required = false, descr = "json of the keys to fetch")
-      val name: ScallopOption[String] = opt[String](required = true, descr = "name of the join/group-by to fetch")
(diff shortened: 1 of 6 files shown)


hanyuli1995 (approved):
LGTM. 👍

spark/src/main/scala/ai/chronon/spark/Driver.scala
```diff
@@ -675,10 +677,17 @@ object Driver {
           if (args.`type`() == "join-stats") {
             fetchStats(args, objectMapper, keyMap, fetcher)
           } else {
+            val featureName = if (args.name.isDefined) {
+              args.name()
+            } else {
+              args.confPath().split("/").takeRight(3).mkString("/")
```
donghanz, line 683: `takeRight(3)` might be error-prone if we update the config path. how about putting this logic in a centralized place like a helper method?

better365: refactored and done.

PR #756, opened 2024-05-01
Update Analyzer code to catch the table permission issue

hanyuli1995:
## Summary

We have found cases where the analyzer could not catch the table permission issue. This PR it trying to fix the issue. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested
Tested with Airbnb internal trust join
```
[2024-05-01 00:13:10] {Analyzer:351} - ---- Table permission check completed. Found permission errors in 1 tables ----
[2024-05-01 00:13:10] {Analyzer:352} - trust.no_access_table
```

## Checklist
- [ ] Documentation update

## Reviewers
@SophieYu41

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
@@ -250,7 +250,7 @@ case class TableUtils(sparkSession: SparkSession) {
       sparkSession.sql(s"SELECT * FROM $tableName where $partitionColumn='$partitionFilter' LIMIT 1").collect()
       true
     } catch {
-      case e: RuntimeException =>
+      case e: Exception =>
         if (e.getMessage.contains("ACCESS DENIED"))
           logger.error(s"[Error] No access to table: $tableName ")
         else {
@@ -258,10 +258,6 @@ case class TableUtils(sparkSession: SparkSession) {
           e.printStackTrace()
         }
         false
-      case ex: Exception =>
-        logger.error(s"[Error] Encountered exception when reading table: $tableName.")
-        ex.printStackTrace()
-        true
     }
   }
 
--- a/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
@@ -17,7 +17,6 @@
 package ai.chronon.spark
 
 import java.io.{PrintWriter, StringWriter}
-
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -250,18 +250,14 @@ case class TableUtils(sparkSession: SparkSession) {
...
-      case e: RuntimeException =>
+      case e: Exception =>
         if (e.getMessage.contains("ACCESS DENIED"))
           logger.error(s"[Error] No access to table: $tableName ")
         else {
           logger.error(s"[Error] Encountered exception when reading table: $tableName.")
           e.printStackTrace()
         }
         false
-      case ex: Exception =>
```
SophieYu41, line 261: Ideally we should figure out why the error message does not fall into first case "ACCESS DENIED" since it do have the target string, and handle it properly. 

Remove the last case would make permission check more aggressive - any exception we encountered would result in a "No permission" which might not be accurate.

hanyuli1995: @SophieYu41 I updated the code. The issue is because the exception is recognized as sparkException instead of java runtime exception, it does include "ACCESS DENIED" in error message. 

One more question, in which case we still want the analyzer succeed with exception caught here? No matter what type of exception it is we still need to remind users reading from tables failed right?

SophieYu41: Great! Thanks!

> in which case we still want the analyzer succeed with exception caught here?
Think it is about unknown of the unknowns.  In case there is some unexpected exception/runtimeException stopping us evaluating the table permission, the question is if we would like to hard-stop the backfill. 
I would prefer to not since real permission issue would surface during backfill anyways, the check here is to fail fast. But if we have false negative, backfill will fail even there is no permission issue. Feels like less harmful to have false-positive rather than false negative.

nikhilsimha: We will hard stop regardless no?

SophieYu41: @nikhilsimha we would hard stop if any validations failed. Think the questions here is - do we want to hard stop if we encountered errors and not able to verify if table permission is present or not.

SophieYu41 (approved):
Thanks for the fix!

better365 (approved):

PR #759, opened 2024-05-03
feat: support overrides

donghanz:
## Summary

feat: support overrides
- support overrides for AirflowAdapter, allowing custom logic to connect to Airflow and trigger a DAG run

## Test Plan

- [x] tested with Airbnb airflow adapter

## Reviewers
@hanyuli1995 @nikhilsimha @ezvz

### DIFF
--- a/api/py/ai/chronon/scheduler/adapters/airflow_adapter.py
+++ b/api/py/ai/chronon/scheduler/adapters/airflow_adapter.py
@@ -1,16 +1,13 @@
 from datetime import datetime
 
-import airflow_client
 from ai.chronon.scheduler.interfaces.orchestrator import WorkflowOrchestrator
 
 from airflow import DAG
 from airflow.operators.bash_operator import BashOperator
 
-AIRFLOW_CLUSTER = airflow_client.Service.STONE
-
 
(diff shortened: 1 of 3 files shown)


better365 (approved):

api/py/ai/chronon/scheduler/adapters/airflow_adapter.py
```diff
@@ -19,7 +16,7 @@ def __init__(self, dag_id, start_date, schedule_interval="@once", airflow_cluste
         self.airflow_cluster = airflow_cluster
 
     def setup(self):
-        airflow_client.init(self.airflow_cluster)
+        """Initialize a connection to Airflow"""
```
ezvz, line 19: Did you mean to delete the airflow_client code here? Do you want to log the string?

ezvz (approved):

api/py/ai/chronon/scheduler/adapters/airflow_adapter.py
donghanz: We will move `airflow_client` import internal. Then this `airflow_adapter` can be used for both adhoc runs (triggered by airflow_client) and production runs (scheduled by Airflow)

PR #760, opened 2024-05-06
Build metadata endpoint and directory walker

hanyuli1995:
## Summary

We are supporting metadata upload to k-v store for key-value pair key->conf right now. We want to add a general class metadata endpoint to support more potential use cases.

This PR is to add two general class **MetadataEndPoint** and **MetadataDirWalker**

MetadataEndPoint: 
```
case class MetadataEndPoint[Conf <: TBase[_, _]: Manifest: ClassTag](
    extractFn: (String, Conf) => (String, String),
    name: String
)
```

Defined with a extract function and an end point name. Extract function extracts the key-value pair from Conf(could be Join/GroupBy/StagingQuery) and file path(string). The name is the dataset name when we send the data to k-v store. 

MetadataDirWalker:
```
class MetadataDirWalker(dirPath: String, metadataEndPointNames: List[String])
```

Go through the directory to iterate over all the config files and generate k-v pair metadata based on the metadata end points provided.

The PR adds two metadata endpoint ZIPLINE_METADATA and ZIPLINE_METADATA_BY_TEAM 

CHRONON_METADATA:
key -> conf json in string format e.g : joins/team/team.example_join.v1 -> {...}

CHRONON_METADATA_BY_TEAM:
type/team -> list of key in string format e.g : joins/team -> a, b, c

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

Testing by running the metadata-upload and fetch command for a join. 
https://docs.google.com/document/d/1X7n_jskS7JyiiqVB3pStgPilg6luy23ho58twCHeip8/edit?usp=sharing

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -669,12 +669,13 @@ object Extensions {
       filePath.split("/").takeRight(3).mkString("/")
     }
 
-    def confPathToTeam: Option[String] = {
+    // derive a feature team key from config, ex: joins/team
+    def confPathToTeamKey: Option[String] = {
       try {
         filePath match {
-          case value if value.contains("staging_queries/") => loadJsonToConf[StagingQuery](value).map(_.metaData.team)
-          case value if value.contains("joins/")           => loadJsonToConf[Join](value).map(_.metaData.team)
-          case value if value.contains("group_bys/")       => loadJsonToConf[GroupBy](value).map(_.metaData.team)
+          case value if value.contains("staging_queries/") => loadJsonToConf[StagingQuery](value).map("staging_queries" + _.metaData.team)
+          case value if value.contains("joins/")           => loadJsonToConf[Join](value).map("joins" + _.metaData.team)
+          case value if value.contains("group_bys/")       => loadJsonToConf[GroupBy](value).map("group_bys" + _.metaData.team)
           case _                                           => logger.info(s"unknown config type in file $filePath"); None
         }
       } catch {
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -640,8 +640,6 @@ object Extensions {
 
   implicit class filePathOps(filePath: String) {
     @transient lazy val logger = LoggerFactory.getLogger(getClass)
-    // process chronon configs only. others will be ignored
-    // todo: add metrics
 
     private def loadJsonToConf[T <: TBase[_, _]: Manifest: ClassTag](file: String): Option[T] = {
       try {
@@ -667,22 +665,6 @@ object Extensions {
     // capture <conf_type>/<team>/<conf_name> as key e.g joins/team/team.example_join.v1
     def confPathToKey: String = {
       filePath.split("/").takeRight(3).mkString("/")
-    }
-
-    // derive a feature team key from config, e.g: joins/team
-    def confPathToTeamKey: Option[String] = {
-      try {
-        filePath match {
-          case value if value.contains("staging_queries/") => loadJsonToConf[StagingQuery](value).map("staging_queries/" + _.metaData.team)
-          case value if value.contains("joins/")           => loadJsonToConf[Join](value).map("joins/" + _.metaData.team)
-          case value if value.contains("group_bys/")       => loadJsonToConf[GroupBy](value).map("group_bys/" + _.metaData.team)
-          case _                                           => logger.info(s"unknown config type in file $filePath"); None
-        }
-      } catch {
-        case e: Throwable =>
-          logger.error(s"Failed to parse compiled team from file path: $filePath, \nerror=${e.getMessage}")
-          None
-      }
     }
 
     def confPathToOptConfStr: Option[String] = {
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -23,13 +23,11 @@ import com.fasterxml.jackson.core.`type`.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.apache.spark.sql.Column
 import org.apache.spark.sql.functions.expr
-import org.apache.thrift.TBase
 
 import java.io.{PrintWriter, StringWriter}
 import java.util
 import java.util.regex.Pattern
 import scala.collection.{Seq, mutable}
-import scala.reflect.ClassTag
 import scala.util.ScalaJavaConversions.{IteratorOps, ListOps, MapOps}
 import scala.util.{Failure, Success, Try}
 
@@ -636,45 +634,11 @@ object Extensions {
     def sanitize: String = Option(string).map(_.replaceAll("[^a-zA-Z0-9_]", "_")).orNull
 
     def cleanSpec: String = string.split("/").head
-  }
 
-  implicit class filePathOps(filePath: String) {
-    @transient lazy val logger = LoggerFactory.getLogger(getClass)
-
-    private def loadJsonToConf[T <: TBase[_, _]: Manifest: ClassTag](file: String): Option[T] = {
-      try {
-        val configConf = ThriftJsonCodec.fromJsonFile[T](file, check = true)
-        Some(configConf)
-      } catch {
-        case e: Throwable =>
-          logger.error(s"Failed to parse compiled Chronon config file: $file, \nerror=${e.getMessage}")
-          None
-      }
-    }
-    private def loadJson[T <: TBase[_, _]: Manifest: ClassTag](file: String): Option[String] = {
-      try {
-        val configConf = loadJsonToConf[T](file).get
-        Some(ThriftJsonCodec.toJsonStr(configConf))
-      } catch {
-        case e: Throwable =>
-          logger.error(s"Failed to parse compiled Chronon config file: $file, \nerror=${e.getMessage}")
-          None
-      }
-    }
     // derive a feature name key from path to file
-    // capture <conf_type>/<team>/<conf_name> as key e.g joins/team/team.example_join.v1
     def confPathToKey: String = {
-      filePath.split("/").takeRight(3).mkString("/")
-    }
-
-    def confPathToOptConfStr: Option[String] = {
-      val confJsonOpt = filePath match {
-        case value if value.contains("staging_queries/") => loadJson[StagingQuery](value)
-        case value if value.contains("joins/")           => loadJson[Join](value)
-        case value if value.contains("group_bys/")       => loadJson[GroupBy](value)
-        case _                                           => logger.info(s"unknown config type in file $filePath"); None
-      }
-      confJsonOpt
+      // capture <conf_type>/<team>/<conf_name> as key e.g joins/team/team.example_join.v1
+      string.split("/").takeRight(3).mkString("/")
     }
   }
 
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -673,9 +673,9 @@ object Extensions {
     def confPathToTeamKey: Option[String] = {
       try {
         filePath match {
-          case value if value.contains("staging_queries/") => loadJsonToConf[StagingQuery](value).map("staging_queries" + _.metaData.team)
-          case value if value.contains("joins/")           => loadJsonToConf[Join](value).map("joins" + _.metaData.team)
-          case value if value.contains("group_bys/")       => loadJsonToConf[GroupBy](value).map("group_bys" + _.metaData.team)
+          case value if value.contains("staging_queries/") => loadJsonToConf[StagingQuery](value).map("staging_queries/" + _.metaData.team)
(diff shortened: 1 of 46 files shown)


spark/src/main/scala/ai/chronon/spark/Driver.scala
```diff
@@ -732,8 +732,16 @@ object Driver {
     }
 
     def run(args: Args): Unit = {
-      val putRequest = args.metaDataStore.putConf(args.confPath())
-      val res = Await.result(putRequest, 1.hour)
+      val acceptedEndPoints = List("ZIPLINE_METADATA", "ZIPLINE_METADATA_BY_TEAM")
```
hanyuli1995, line 735: We probably can make this a passed-in config, WDYT? @nikhilsimha

online/src/main/scala/ai/chronon/online/MetadataStore.scala
```diff
@@ -188,6 +182,34 @@ class MetadataStore(kvStore: KVStore, val dataset: String = ChrononMetadataKey,
...
+             |key: $k
+             |conf: $v""".stripMargin)
+        val kBytes = k.getBytes()
+        val vBytes =
+          if (v.isEmpty) Array.emptyByteArray
+          else if (v.length == 1) v.head.getBytes()
+          else v.mkString("\n").getBytes()
+        PutRequest(keyBytes = kBytes,
+                   valueBytes = vBytes,
+                   dataset = dataset,
```
hanyuli1995, line 203: I am using the end point name as dataset here.

api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -634,11 +636,63 @@ object Extensions {
...
     // derive a feature name key from path to file
+    // capture <conf_type>/<team>/<conf_name> as key e.g joins/team/team.example_join.v1
     def confPathToKey: String = {
-      // capture <conf_type>/<team>/<conf_name> as key e.g joins/team/team.example_join.v1
-      string.split("/").takeRight(3).mkString("/")
+      filePath.split("/").takeRight(3).mkString("/")
+    }
+
+    // derive a feature team key from config, e.g: joins/team
+    def confPathToTeamKey: Option[String] = {
```
nikhilsimha, line 673: So we should make it so that we only parse the file once and run a bunch of extractors for the conf type.

So we need to maintain a list of (key, value) extractors for groupBys, joins and stagingQueries.

We simply parse once, run all the extractors per file, print it to console and write to metastore.

spark/src/main/scala/ai/chronon/spark/Driver.scala
nikhilsimha: agreed - we could take this from conf

online/src/main/scala/ai/chronon/online/MetadataDirWalker.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  private def parseName(path: String): Option[String] = {
+    val gson = new Gson()
+    val reader = Files.newBufferedReader(Paths.get(path))
+    try {
+      val map = gson.fromJson(reader, classOf[java.util.Map[String, AnyRef]])
+      Option(map.get("metaData"))
+        .map(_.asInstanceOf[java.util.Map[String, AnyRef]])
+        .map(_.get("name"))
+        .flatMap(Option(_))
```
nikhilsimha, line 34:
```suggestion
        .flatMap(Option(_.get("name")))
```
should work right?

nikhilsimha (commented):
Added a high level comment

api/src/main/scala/ai/chronon/api/Extensions.scala
hanyuli1995: Hi @nikhilsimha this function is only called in the defined extract function in metadata end point. The metadata dir walker will go through the files like you mentioned here. I explained the logic below in metadata DirWalker.

online/src/main/scala/ai/chronon/online/MetadataDirWalker.scala
```diff
@@ -0,0 +1,85 @@
...
+    val (key, value) = endPoint.extractFn(filePath)
+    (key, value)
+  }
+
+  def run: Map[String, Map[String, List[String]]] = {
+    nonEmptyFileList.foldLeft(Map.empty[String, Map[String, List[String]]]) { (acc, file) =>
+      val kvPairToEndPoint: List[(String, (Option[String], Option[String]))] = metadataEndPointNames.map {
+        metadataEndPointName =>
+          (metadataEndPointName, extractKVPair(metadataEndPointName, file.getPath))
+      }
```
hanyuli1995, line 71: In this map function for a file, we iterate over all the end points and call the extract function to generate the key-value pair.

online/src/main/scala/ai/chronon/online/MetadataDirWalker.scala
```diff
@@ -0,0 +1,130 @@
...
+        name.isDefined
+      }
+  }
+
+  /**
+    * Iterate over the list of files and extract the key value pairs for each file
+    * @return Map of endpoint -> (Map of key -> List of values)
+    *         e.g. (
+    *            ZIPLINE_METADATA_BY_TEAM -> (team -> List("join1", "join2")),
+    *            ZIPLINE_METADATA -> (teams/joins/join1 -> config1)
```
better365, line 78:
```suggestion
    *            CHRONON_METADATA_BY_TEAM -> (team -> List("join1", "join2")),
    *            CHRONON_METADATA -> (teams/joins/join1 -> config1)
```

online/src/main/scala/ai/chronon/online/MetadataEndPoint.scala
```diff
@@ -0,0 +1,56 @@
...
+
+case class MetadataEndPoint[Conf <: TBase[_, _]: Manifest: ClassTag](
+    extractFn: (String, Conf) => (String, String),
+    name: String
+)
+object MetadataEndPoint {
+  @transient implicit lazy val logger = LoggerFactory.getLogger(getClass)
+
+  val ConfByKeyEndPointName = "ZIPLINE_METADATA"
+  val NameByTeamEndPointName = "ZIPLINE_METADATA_BY_TEAM"
```
better365, line 18:
```suggestion
  val ConfByKeyEndPointName = "CHRONON_METADATA"
  val NameByTeamEndPointName = "CHRONON_METADATA_BY_TEAM"
```

spark/src/main/scala/ai/chronon/spark/Driver.scala
```diff
@@ -732,8 +732,14 @@ object Driver {
     }
 
     def run(args: Args): Unit = {
-      val putRequest = args.metaDataStore.putConf(args.confPath())
-      val res = Await.result(putRequest, 1.hour)
+      val acceptedEndPoints = List("ZIPLINE_METADATA")
```
better365, line 735: Let's not use the old term `zipline` as much as we could. 
Could you make those configurable or read from a config file or read from command line params? 
```suggestion
      val acceptedEndPoints = List("CHRONON_METADATA")
```

online/src/main/scala/ai/chronon/online/MetadataStore.scala
```diff
@@ -39,7 +38,7 @@ case class DataMetrics(series: Seq[(Long, SortedMap[String, Any])])
 class MetadataStore(kvStore: KVStore, val dataset: String = ChrononMetadataKey, timeoutMillis: Long) {
   @transient implicit lazy val logger = LoggerFactory.getLogger(getClass)
   private var partitionSpec = PartitionSpec(format = "yyyy-MM-dd", spanMillis = WindowUtils.Day.millis)
-  private val CONF_BATCH_SIZE = 50
+  private val CONF_BATCH_SIZE = 100
```
better365, line 41: we could make this configurable. 
We had some put request batch oversize error from our KV store before. That's why we set it to 50 for the safer side.

online/src/main/scala/ai/chronon/online/MetadataStore.scala
```diff
@@ -149,88 +148,31 @@ class MetadataStore(kvStore: KVStore, val dataset: String = ChrononMetadataKey,
...
+      case (k, v) => {
+        logger.info(s"""Putting metadata for
+             |dataset: $datasetName
+             |key: $k
+             |conf: $v""".stripMargin)
+        val kBytes = k.getBytes()
+        val vBytes =
+          if (v.isEmpty) Array.emptyByteArray
+          else if (v.length == 1) v.head.getBytes()
+          else v.mkString("\n").getBytes()
```
better365, line 166: why do we want this? If it is already bytes, what is the use of `\n`?

hanyuli1995: @better365 after the change the value could be a list of string. not sure what will be the value in mussel if apply getBytes on a list instead of a string. I am trying to convert the list to a string first then apply toBytes

online/src/main/scala/ai/chronon/online/MetadataStore.scala
hanyuli1995: Yep it is a configurable parameter in put function. https://github.com/airbnb/chronon/blob/19664365ace064476d49b6830c980612fa2627d5/online/src/main/scala/ai/chronon/online/MetadataStore.scala#L154c I increase this to 100 because 50 doesn't work well for groupBy(there are about 1k groupBy and 200 joins. )

online/src/main/scala/ai/chronon/online/MetadataStore.scala
better365: I see. What will be used for the list of the configs?  it will be easier to code it out here.

better365 (approved):
Stamped. Please fix the failed UT.

online/src/main/scala/ai/chronon/online/MetadataStore.scala
```diff
@@ -149,88 +148,31 @@ class MetadataStore(kvStore: KVStore, val dataset: String = ChrononMetadataKey,
...
+      case (k, v) => {
+        logger.info(s"""Putting metadata for
+             |dataset: $datasetName
+             |key: $k
+             |conf: $v""".stripMargin)
+        val kBytes = k.getBytes()
+        val vBytes =
+          if (v.isEmpty) Array.emptyByteArray
+          else if (v.length == 1) v.head.getBytes()
+          else v.mkString(",").getBytes()
```
nikhilsimha, line 166: this will break if there are commas in a single member string of the seq.

Chatgpt gave me this

```scala
import java.util.Base64
import java.nio.charset.StandardCharsets

object StringArrayConverter {
  // Method to convert an array of strings to a byte array using Base64 encoding for each element
  def stringsToBytes(strings: Array[String]): Array[Byte] = {
    val base64EncodedStrings = strings.map(s => Base64.getEncoder.encodeToString(s.getBytes(StandardCharsets.UTF_8)))
    base64EncodedStrings.mkString("\n").getBytes(StandardCharsets.UTF_8)
  }

  // Method to convert a byte array back to an array of strings by decoding Base64
  def bytesToStrings(bytes: Array[Byte]): Array[String] = {
    val encodedString = new String(bytes, StandardCharsets.UTF_8)
    encodedString.split("\n").map(s => new String(Base64.getDecoder.decode(s), StandardCharsets.UTF_8))
  }
}

// Testing the methods
val originalStrings = Array("a", "b,c,d", "e.f.g", "h|i|j")
val bytes = StringArrayConverter.stringsToBytes(originalStrings)
val decodedStrings = StringArrayConverter.bytesToStrings(bytes)

println("Original Strings: " + originalStrings.mkString("[", ", ", "]"))
println("Decoded Strings: " + decodedStrings.mkString("[", ", ", "]"))
```

nikhilsimha: use "," as you were using instead of the newline here

spark/src/test/scala/ai/chronon/spark/test/FetcherTest.scala
```diff
@@ -74,7 +74,10 @@ class FetcherTest extends TestCase {
     val singleFileMetadataStore = new MetadataStore(inMemoryKvStore, singleFileDataSet, timeoutMillis = 10000)
     inMemoryKvStore.create(singleFileDataSet)
     // set the working directory to /chronon instead of $MODULE_DIR in configuration if Intellij fails testing
-    val singleFilePut = singleFileMetadataStore.putConf(confResource.getPath)
+    val acceptedEndPoints = List(MetadataEndPoint.ConfByKeyEndPointName)
```
nikhilsimha, line 77: I think we should break it out into its own test. There is a metadataexportertest.scala file

online/src/main/scala/ai/chronon/online/MetadataDirWalker.scala
nikhilsimha: got it!

spark/src/test/scala/ai/chronon/spark/test/FetcherTest.scala
hanyuli1995: Hi @nikhilsimha I need to update the tests here because the original putConf method has been removed from metadataStore file.

online/src/main/scala/ai/chronon/online/MetadataStore.scala
hanyuli1995: done

PR #767, opened 2024-05-24
Fix joins to include partial null results

jbrooks-stripe:
## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update


nikhilsimha (approved):
Looks great! Thanks for the fix! There are some UT errors.

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
```diff
@@ -1298,4 +1300,110 @@ class JoinTest {
...
+        ),
+        metaData = Builders.MetaData(
+          name = "test.user_features",
+          namespace = namespace,
+          team = "chronon"
+        )
+      )
+    }
+
+    def assertJoin(columnNames: Seq[String]) = {
```
varant-zlai, line 1386: It looks like `assertJoin` isn't actually called. Do you want to call it in the above function?

PR #777, opened 2024-06-25
Validator/Compiler logic refinement

yuli-han:
## Summary

In compiler we are validating the derivation logic defined by users, and also print out the feature name list if --feature-display flag is enabled. 

There are some bugs in the existing derivation validation and feature print out. This PR is to fix these bugs.

Derivation validation logic:

What can be used as input to derivations (online)
All join_parts features + all external_part value fields 
ds + ts (special handling)

What can be used as input to derivations (offline)
All join_parts features + left source keys + all external_part value fields 
ds + ts (special handling)

In derivation validation, we are validating the renaming derivations, the original column should exist in pre-derived columns(for online case keys are not included, for offline case keys are included). We also validate there is no duplicate columns name.
## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

Tested by running the compiler on a config with derivation defined on external column name: 
Test result for Airbnb users: https://docs.google.com/document/d/12FDawWrC-5QMY70Sx7UguO74QnhZxRZbYJcwlYeiJ2o/edit?usp=sharing

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @hzding621 @donghanz


api/py/ai/chronon/repo/validator.py
```diff
@@ -110,7 +109,29 @@ def get_pre_derived_join_columns(join: Join) -> List[str]:
             prefix = jp.prefix + "_" if jp.prefix else ""
             gb_prefix = jp.groupBy.metaData.name.replace(".", "_")
             output_columns.append(prefix + gb_prefix + "_" + col)
-    return output_columns
+    return output_columns + get_external_columns(join)
```
hzding621, line 112: nit: refactor the output_columns into a separate method e.g. `get_all_group_by_columns` and change this into

```
return get_all_group_by_columns(join) + get_all_external_columns(join)
```

hzding621, line 112: Do we ever print out all available column names when there is a validation failure? It would be nice to have it but doesn't have to be in this PR

hzding621 (approved):

api/py/ai/chronon/repo/validator.py
```diff
@@ -87,52 +103,87 @@ def get_group_by_output_columns(group_by: GroupBy) -> List[str]:
...
+def get_external_part_full_name(external_part: ExternalPart) -> str:
+    prefix = external_part.prefix + "_" if external_part.prefix else ""
+    name = external_part.source.metadata.name
+    sanitized_name = re.sub("[^a-zA-Z0-9_]", "_", name)
+    return "ext_" + prefix + sanitized_name
+
+
+# The external columns name logic should be consistent with the logic defined in fetcher.scala
+# https://github.com/airbnb/chronon/blob/main/online/src/main/scala/ai/chronon/online/Fetcher.scala#L371
+def get_external_columns(join: Join) -> List[str]:
```
hzding621, line 147: nit: get_external_features

api/py/ai/chronon/repo/validator.py
```diff
@@ -284,7 +341,12 @@ def _validate_join(self, join: Join) -> List[str]:
         # Only validate the join derivation when the underlying groupBy is valid
         group_by_correct = all(not errors for errors in group_by_errors)
         if join.derivations and group_by_correct:
-            columns = set(get_pre_derived_join_columns(join))
+            columns = get_pre_derived_join_columns(join)
```
hzding621, line 344: nit: put this in else condition to avoid repeated calculation, because most joins are online

api/py/ai/chronon/repo/validator.py
```diff
@@ -322,7 +384,10 @@ def _validate_group_by(self, group_by: GroupBy) -> List[str]:
 
         # validate the derivations are defined correctly
         if group_by.derivations:
-            columns = set(get_pre_derived_group_by_columns(group_by))
+            columns = get_pre_derived_group_by_columns(group_by)
```
hzding621, line 387: nit: put this in else condition to avoid repeated calculation, because most gb are online

hzding621:
@yuli-han Can we also improve the unit tests?

api/py/ai/chronon/repo/validator.py
yuli-han: One question here is, external part is purely an online concept right? And since we don't include external part keys for online case does it mean external_features = external_columns?

api/py/ai/chronon/repo/validator.py
yuli-han: updated

api/py/ai/chronon/repo/validator.py
yuli-han: updated

api/py/ai/chronon/repo/validator.py
yuli-han: updated

api/py/ai/chronon/repo/validator.py
```diff
@@ -87,52 +103,87 @@ def get_group_by_output_columns(group_by: GroupBy) -> List[str]:
...
-
-    return list(output_columns)
+        return build_derived_columns(output_columns, group_by.derivations)
+    else:
+        return list(output_columns)
 
 
-def get_pre_derived_join_columns(join: Join) -> List[str]:
-    output_columns = []
+def get_pre_derived_join_features(join: Join) -> List[str]:
```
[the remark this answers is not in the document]
yuli-han: updated

api/py/ai/chronon/repo/validator.py
```diff
@@ -87,52 +103,87 @@ def get_group_by_output_columns(group_by: GroupBy) -> List[str]:
...
-def get_join_output_columns(join: Join) -> List[str]:
+def get_pre_derived_join_columns(join: Join) -> List[str]:
+    internal_columns = []
+    if join.left.events:
+        internal_columns.extend(join.left.events.query.selects.keys())
+    elif join.left.entities:
+        internal_columns.extend(join.left.entities.query.selects.keys())
+    elif join.left.joinSource:
+        internal_columns.extend(join.left.joinSource.query.selects.keys())
+    internal_columns.extend(get_pre_derived_join_features(join))
```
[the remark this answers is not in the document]
yuli-han: Updated

hzding621 (approved):

api/py/ai/chronon/repo/validator.py
```diff
@@ -87,52 +103,88 @@ def get_group_by_output_columns(group_by: GroupBy) -> List[str]:
...
             prefix = jp.prefix + "_" if jp.prefix else ""
             gb_prefix = jp.groupBy.metaData.name.replace(".", "_")
-            output_columns.append(prefix + gb_prefix + "_" + col)
-    return output_columns
+            internal_features.append(prefix + gb_prefix + "_" + col)
+    return internal_features
 
 
-def get_join_output_columns(join: Join) -> List[str]:
+def get_pre_derived_join_keys(join: Join) -> List[str]:
```
donghanz, line 124: this function can be more generalized like 
```
def get_pre_derived_join_keys(source: Source) -> List[str]:
```
Then it can be used [here](https://github.com/airbnb/chronon/pull/777/files#diff-ebaa3fcfb1f15efb4e35e094b3a6dea5d8f01120a3ded8ab53d58fbebf9c0078R86-R91) to avoid duplicate logic

donghanz, line 124: this function can be more generalized like 
```
def get_pre_derived_join_keys(source: Source) -> List[str]:
```
Then it can be used [here](https://github.com/airbnb/chronon/pull/777/files#diff-ebaa3fcfb1f15efb4e35e094b3a6dea5d8f01120a3ded8ab53d58fbebf9c0078R86-R91) to avoid duplicate logic

api/py/ai/chronon/repo/validator.py
```diff
@@ -74,11 +74,27 @@ def is_identifier(s: str) -> bool:
...
             output_columns.extend(get_output_col_names(agg))
+    # For group_bys without aggregations, selected fields from query
+    else:
+        for source in group_by.sources:
+            if source.events:
+                output_columns.extend(source.events.query.selects.keys())
+            elif source.entities:
+                output_columns.extend(source.entities.query.selects.keys())
+            elif source.joinSource:
+                output_columns.extend(source.joinSource.query.selects.keys())
```
donghanz, line 91: as mentioned below, it can use `get_pre_derived_join_keys()` in this way:
```
for source in group_by.sources:
    output_columns.extend(get_pre_derived_join_keys(source))
```

api/py/ai/chronon/repo/validator.py
```diff
@@ -87,52 +103,88 @@ def get_group_by_output_columns(group_by: GroupBy) -> List[str]:
...
+        return list(join.left.events.query.selects.keys())
+    elif join.left.entities:
+        return list(join.left.entities.query.selects.keys())
+    elif join.left.joinSource:
+        return list(join.left.joinSource.query.selects.keys())
+
+
+# The logic should be consistent with the full name logic defined
+# in https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Extensions.scala#L677.
+def get_external_part_full_name(external_part: ExternalPart) -> str:
```
donghanz, line 135: why don't we call `external_part.full_name` directly?

donghanz, line 135: how difficulty if we want to add a Python equivalent of `externalPart.fullName` as in Scala?

yuli-han: The struct is defined in thrift. I think we can add it but probably don't want to introduce too much changes in this one single bug fix PR.

api/py/ai/chronon/repo/validator.py
yuli-han: Updated

api/py/ai/chronon/repo/validator.py
yuli-han: Updated

donghanz (approved):

PR #779, opened 2024-06-26
fix: Update derivation response map to pass through base exceptions 

hzding621:
## Summary

One of the UTs of `FetcherTest` is failing on main branch. The RC is that we updated fetcher response map logic when there is failure in derivations. 

This PR updated the logic to pass through the base exceptions. Base map fetching failure will **guarantee** failure of derivation because some fields are missing, so passing through the base exception is important to keep trace of the original issue.  Also the UT is updated to match the new behavior. 

## Why / Goal

Fix a UT on main branch, and properly handle exception logging in response map.

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@pengyu-hou @yuli-han @donghanz

### DIFF
--- a/online/src/main/scala/ai/chronon/online/Fetcher.scala
+++ b/online/src/main/scala/ai/chronon/online/Fetcher.scala
@@ -250,10 +250,13 @@ class Fetcher(val kvStore: KVStore,
                       Map("derivation_fetch_exception" -> exception.traceString.asInstanceOf[AnyRef])
                     renameOnlyDerivedMap ++ derivedExceptionMap
                 }
+                // Preserve exceptions from baseMap
+                val baseMapExceptions = baseMap.filter(_._1.endsWith("_exception"))
+                val finalizedDerivedMap = derivedMap ++ baseMapExceptions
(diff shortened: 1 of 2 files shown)


yuli-han (approved):
Thanks for catching the issue!

donghanz (approved):

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -250,10 +250,13 @@ class Fetcher(val kvStore: KVStore,
                       Map("derivation_fetch_exception" -> exception.traceString.asInstanceOf[AnyRef])
                     renameOnlyDerivedMap ++ derivedExceptionMap
                 }
+                // Preserve exceptions from baseMap
+                val baseMapExceptions = baseMap.filter(_._1.endsWith("_exception"))
```
pengyu-hou, line 254: @yuli-han will we have the similar issue for fetch group by?

PR #780, opened 2024-06-27
fix: analyzer handle multiple sources and bootstrap 

hzding621:
## Summary

Update analyzer logic: 
- handle multiple sources
- handle bootstrap (skip validation)

## Why / Goal

GroupBy with multiple sources is supported and all sources is "unioned" before processing the group by. So the data availability check in analyer only need to verify that the combined date range of the sources need to cover required date ranges. Current logic requires each source covering all date range which is wrong. 

For joins with bootstrap, there is no straightforward to identify the required date range without actually running the bootstrap. So for joins with bootstrap we simply don't raise data availability "errors" (but only print them). 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested 

Ran local JAR on user joins 

## Checklist
- [ ] Documentation update

## Reviewers

@yuli-han @pengyu-hou @donghanz

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -352,7 +352,7 @@ class Analyzer(tableUtils: TableUtils,
       logger.info(noAccessTables.mkString("\n"))
       logger.info(s"---- Data availability check completed. Found issue in ${dataAvailabilityErrors.size} tables ----")
       dataAvailabilityErrors.foreach(error =>
-        logger.info(s"Table ${error._1} : Group_By ${error._2} : Expected start ${error._3}"))
+        logger.info(s"Group_By ${error._2} : Source Tables ${error._1} : Expected start ${error._3}"))
     }
 
     if (validationAssert) {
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -430,21 +439,36 @@ class Analyzer(tableUtils: TableUtils,
...
-            logger.info(s"Checking table $table for data availability ... Expected start partition: $expectedStart")
-            val existingPartitions = tableUtils.partitions(table)
-            val minPartition = existingPartitions.reduceOption((a, b) => Ordering[String].min(a, b))
-            //check if partition available or table is cumulative
-            if (!tableUtils.ifPartitionExistsInTable(table, expectedStart) && !source.isCumulative) {
-              println(s"""
-                         |Join needs data older than what is available for GroupBy: ${groupBy.metaData.name}, table: $table,
+          logger.info(
+            s"Checking data availability for group_by ${groupBy.metaData.name} ... Expected start partition: $expectedStart")
+          if (groupBy.sources.toScala.exists(s => s.isCumulative)) {
```
SophieYu41, line 444: 👍

SophieYu41 (approved):
Thanks for the fix!

donghanz (approved):
Thanks!

PR #781, opened 2024-07-02
Add bounded unique count aggregation

jbrooks-stripe:
## Summary

Adds a BOUNDED_UNIQUE_COUNT aggregation. This will allow exact unique/distinct counts, but will cap at a given value to keep memory usage constant.

## Why / Goal
We have use cases where we'd prefer an exact solution instead of the approx equivalents, but want to have protections in place so that memory doesn't become an issue. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [x] Documentation update

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
@@ -635,7 +635,7 @@ class BoundedUniqueCount[T](inputType: DataType, k: Int = 8) extends SimpleAggre
   override def irType: DataType = ListType(StringType)
 
   override def merge(ir1: util.Set[String], ir2: util.Set[String]): util.Set[String] = {
-    ir2.asScala.foreach(v =>
+    ir2.iterator().asScala.foreach(v =>
       if (ir1.size() < k) {
         ir1.add(v)
       })
--- a/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
@@ -600,7 +600,12 @@ class ApproxHistogram[T: FrequentItemsFriendly](mapSize: Int, errorType: ErrorTy
   }
 }
 
-class BoundedUniqueCount[T](inputType: DataType, k: Int = 8) extends SimpleAggregator[T, util.Set[String], Long] {
+object BoundedUniqueCount {
+  private val SentinelSet: util.Set[Any] = new util.HashSet[Any]()
+  private val SentinelMarker: String = "__SENTINEL__"
+}
+
+class BoundedUniqueCount[T](inputType: DataType, k: Int = 8) extends SimpleAggregator[T, util.Set[Any], Long] {
   private def toBytes(input: T): Array[Byte] = {
     val bos = new ByteArrayOutputStream()
     val out = new ObjectOutputStream(bos)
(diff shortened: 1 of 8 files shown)


piyushn-stripe (approved):

hzding621 (approved):

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -599,6 +600,59 @@ class ApproxHistogram[T: FrequentItemsFriendly](mapSize: Int, errorType: ErrorTy
...
+    ir.add(hashInput(input))
+    ir
+  }
+
+  override def outputType: DataType = LongType
+
+  override def irType: DataType = ListType(StringType)
+
+  override def merge(ir1: util.Set[String], ir2: util.Set[String]): util.Set[String] = {
+    ir2.asScala.foreach(v =>
```
nikhil-zlai:
```suggestion
    ir2.iterator().asScala.foreach(v =>
```

nikhil-zlai: otherwise it will create intermediate collections

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -599,6 +600,59 @@ class ApproxHistogram[T: FrequentItemsFriendly](mapSize: Int, errorType: ErrorTy
...
+  private def toBytes(input: T): Array[Byte] = {
+    val bos = new ByteArrayOutputStream()
+    val out = new ObjectOutputStream(bos)
+    out.writeObject(input)
+    out.flush()
+    bos.toByteArray
+  }
+
+  private def md5Hex(bytes: Array[Byte]): String =
+    MessageDigest.getInstance("MD5").digest(bytes).map("%02x".format(_)).mkString
```
nikhil-zlai, line 618: lets say i want to unique count a bunch of user / merchant ids (long values) - won't this be less efficient than simply keeping the set of longs?

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -599,6 +600,59 @@ class ApproxHistogram[T: FrequentItemsFriendly](mapSize: Int, errorType: ErrorTy
...
+
+  override def prepare(input: T): util.Set[String] = {
+    val result = new util.HashSet[String](k)
+    result.add(hashInput(input))
+    result
+  }
+
+  override def update(ir: util.Set[String], input: T): util.Set[String] = {
+    if (ir.size() >= k) {
+      return ir
```
nikhil-zlai: memory optimization: we can use a sentinel set when k is reached.


```suggestion
      if(ir == Constants.SentinelSet || ir.size() >= k) return Constants.SentinelSet
```

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -599,6 +600,59 @@ class ApproxHistogram[T: FrequentItemsFriendly](mapSize: Int, errorType: ErrorTy
...
+  override def merge(ir1: util.Set[String], ir2: util.Set[String]): util.Set[String] = {
+    ir2.asScala.foreach(v =>
+      if (ir1.size() < k) {
+        ir1.add(v)
+      })
+
+    ir1
+  }
+
+  override def finalize(ir: util.Set[String]): Long = ir.size()
```
nikhil-zlai:
```suggestion
  override def finalize(ir: util.Set[String]): Long = if(ir == Constants.SentinelSet) k else ir.size()
```

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
pengyu-hou: Good call out!

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
pengyu-hou: Hm.. I don't think we have sentinel set yet in OSS branch.

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
pengyu-hou: done

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
pengyu-hou: made the code change to keep the numeric type as is

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -599,6 +600,113 @@ class ApproxHistogram[T: FrequentItemsFriendly](mapSize: Int, errorType: ErrorTy
...
+    }
+  }
+
+  override def normalize(ir: util.Set[Any]): Any = {
+    if (ir == BoundedUniqueCount.SentinelSet) {
+      val list = new util.ArrayList[Any]()
+      list.add(BoundedUniqueCount.SentinelMarker)
+      list
+    } else {
+      new util.ArrayList[Any](ir)
```
nikhil-zlai, line 696: aren't both of these same - can we just do


```suggestion
      new util.ArrayList[Any](ir)
```

nikhil-zlai (approved):
This looks great! thanks pengyu!

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
pengyu-hou: Hi @nikhil-zlai , this is used to differentiate the empty Sentinel Set and the actual empty set.

PR #783, opened 2024-07-09
fix: join source backfill

donghanz:
## Summary

- build a join flow in the case of join source
- skip analyzer in join part tasks
- more logging
- bug fix: `None.get`
- skip small mode in the case of join source
- set computeDependency to False if not a mega join
## Why / Goal

Join source backfill couldn't run in parallel as the normal join. They were bundled in the left table computation. With this fix, we put the join source backfill as a upstream. After it is done, the downstream join, which is the actual one that user defines, will be triggered.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested: internal testing

### DIFF
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -444,7 +444,7 @@ object Extensions {
     def semanticHash: String = {
       val newGroupBy = groupBy.deepCopy()
       newGroupBy.unsetMetaData()
-      logger.info(s"GroupBy object: ${ThriftJsonCodec.toJsonStr(newGroupBy)}")
+      logger.info(s"Computing semantic hash for GroupBy object: ${ThriftJsonCodec.toJsonStr(newGroupBy)}")
       ThriftJsonCodec.md5Digest(newGroupBy)
     }
 
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -424,6 +424,7 @@ object Extensions {
   }
 
   implicit class GroupByOps(groupBy: GroupBy) extends GroupBy(groupBy) {
+    @transient lazy val logger = LoggerFactory.getLogger(getClass)
     def maxWindow: Option[Window] = {
       val allWindowsOpt = Option(groupBy.aggregations)
         .flatMap(_.toScala.toSeq.allWindowsOpt)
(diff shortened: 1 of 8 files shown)


spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -418,21 +418,23 @@ abstract class JoinBase(joinConf: api.Join,
...
-    } catch {
-      case ex: AssertionError =>
-        metrics.gauge(Metrics.Name.validationFailure, 1)
-        logger.error(s"Validation failed. Please check the validation error in log.")
-        if (tableUtils.backfillValidationEnforced) throw ex
-      case e: Throwable =>
-        metrics.gauge(Metrics.Name.validationFailure, 1)
-        logger.error(s"An unexpected error occurred during validation. ${e.getMessage}")
+    // Run validations before starting the job; Skip validation if selectedJoinParts is defined
+    if (selectedJoinParts.isEmpty) {
```
pengyu-hou, line 422: I am a little lost here. Should it run analyzer when selected join parts is not empty? 
```suggestion
    if (!selectedJoinParts.isEmpty) {
```

donghanz: When `selectedJoinParts` is not empty, it is a join part task. We don't have to run the analyzer on the join parts as they have been validated in the left table task and will be checked again in the final table task. Hence, we only run the analyzer for non JP use cases - when `selectedJoinParts` is empty.

api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -443,6 +444,7 @@ object Extensions {
     def semanticHash: String = {
       val newGroupBy = groupBy.deepCopy()
       newGroupBy.unsetMetaData()
+      logger.info(s"GroupBy object: ${ThriftJsonCodec.toJsonStr(newGroupBy)}")
```
hzding621, line 447: nit: computing semantic hash for groupBy object:

spark/src/main/scala/ai/chronon/spark/JoinBase.scala
pengyu-hou: if the selected join parts are empty, do we still need to run analyzer? would it be an invalid join?

hzding621 (approved):

spark/src/main/scala/ai/chronon/spark/JoinBase.scala
donghanz: Yes - we do. It is not only for backward compatibility, but also there are valid use cases like [here](https://github.com/airbnb/chronon/blob/97b06ec6c13b2289d86130b1bbb37de9d0aab998/spark/src/main/scala/ai/chronon/spark/stats/ConsistencyJob.scala#L84) without any selected join parts. That's why we didn't wipe out the analyzer from here.

pengyu-hou (approved):

PR #793, opened 2024-07-22
fix: downgrade caffeine to 2.8.5

hzding621:
## Summary

downgrade caffeine from 2.9.3. to 2.8.5 

## Why / Goal

dependency conflict of an upstream library org.checkerframework:checker-qual 3.4.1 vs 3.19.0 used by airbnb

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@caiocamatta-stripe @yuli-han @pengyu-hou

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -310,7 +310,7 @@ lazy val online = project
       "com.datadoghq" % "java-dogstatsd-client" % "2.7",
       "org.rogach" %% "scallop" % "4.0.1",
       "net.jodah" % "typetools" % "0.4.1",
-      "com.github.ben-manes.caffeine" % "caffeine" % "2.9.3"
+      "com.github.ben-manes.caffeine" % "caffeine" % "2.8.5"
     ),
     libraryDependencies ++= fromMatrix(scalaVersion.value, "spark-all", "scala-parallel-collections", "netty-buffer"),
     version := git.versionProperty.value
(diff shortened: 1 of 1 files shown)


pengyu-hou (approved):

pengyu-hou:
@caiocamatta-stripe @piyushn-stripe 
Would this version work for Stripe? We are seeing some internal deps conflicts.

caiocamatta-stripe (commented):
@pengyu-hou @hzding621 

I looked at the release notes for 2.8.5 -> 2.9.3 and I don't see any obvious issues with downgrading Caffeine.  https://github.com/ben-manes/caffeine/releases?page=2

There are a number of of bug fixes, but most (maybe all) of them don't affect the code we have in Chronon. The only change I saw that potentially affect us is this optimization to `putIfAbsent` https://github.com/ben-manes/caffeine/issues/506, but we only use `put()` so it should be OK.

Quick question before I approve: how hard would it be for you to fix the conflict internally?

ben-manes:
What about excluding the annotation dependency? They are not needed at runtime and should be forward compatible if you allow the version upgrade. When appropriate, please consider using v3.x.

hzding621:
@caiocamatta-stripe the conflict we see is from a dependency (`org.checkerframework:checker-qual`) , which is at 3.19.0 for caffeine at 2.9.3. Internally we have a dependency of this lib at 3.4.1. And there is some non-compatible API between 3.19.0 and 3.4.1. 

> Quick question before I approve: how hard would it be for you to fix the conflict internally?

We'd have to version bump quite a few user services that rely on chronon fetcher (imported as libraries).

caiocamatta-stripe (approved):
Alright this looks fine to me. Next time I touch our caching code I'll see if we can upgrade to 3.x and exclude the annotation dependency as Ben suggested

PR #794, opened 2024-07-23
Adding Timestamp Check to GroupBy Analyzer

pkundurthy:
## Summary

- This PR adds checks on timestamps in a GroupBy source 
    1. If it contains all contains NULLs in the timestamp column -> fail
    2. If it contains any timestamps out of epoch miillsec range (1971 - 2099)  -> fail
        (31536000000L (which is milliseconds of 1971-01-01 00:00:00 UTC) to 
         2099-12-31 23:59:59 (4102473599999L)
- The Analyzer class' method `analyzeGroupBy` will run the above checks all the time. 
- Checks will only run when `ts` column is provided. 

## Why / Goal

- For temporal GroupBys if an incorrect timestamp field is provided (for e.g. UNIX_TIMESTAMP(ts_start, 'yyyy-MM-dd') -> NULL), this option will ensure that we catch such cases. 

## Test Plan

- [X] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

Integration test with `CAST(NULL AS BIGINT)` for timestamp field
```
[2024-07-27 02:01:33] {Analyzer:193} - Running GroupBy analysis for group_by/airbnb_org.response_configs.invitations_by_referral.v3 ...
Exception in thread "main" java.lang.AssertionError: assertion failed: [ERROR]: GroupBy validation failed.
 Please check that source has non-null timestamps.
 check notNullCount: 0
 ```
 
 Integration test with `UNIX_TIMESTAMP(updated_at, 'yyyy-MM-dd HH:mm:ss.SSS') * 1000000` for timestamp field 
 ```
 [2024-07-27 02:05:25] {Analyzer:193} - Running GroupBy analysis for group_by/airbnb_org.response_configs.invitations_by_referral.v3 ...
Exception in thread "main" java.lang.AssertionError: assertion failed: [ERROR]: GroupBy validation failed.
 Please check that source has valid epoch millisecond timestamps.
 badRangeCount: 127
 ```

## Checklist
- [ ] Documentation update

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -187,8 +187,7 @@ class Analyzer(tableUtils: TableUtils,
                      prefix: String = "",
                      includeOutputTableName: Boolean = false,
                      enableHitter: Boolean = false,
-                     validationAssert: Boolean = false
-                    ): (Array[AggregationMetadata], Map[String, DataType]) = {
+                     validationAssert: Boolean = false): (Array[AggregationMetadata], Map[String, DataType]) = {
     groupByConf.setups.foreach(tableUtils.sql)
     val groupBy = GroupBy.from(groupByConf, range, tableUtils, computeDependency = enableHitter, finalize = true)
     val name = "group_by/" + prefix + groupByConf.metaData.name
@@ -244,7 +243,8 @@ class Analyzer(tableUtils: TableUtils,
     }
 
     if (validationAssert) {
-      assert(checkTs == "Acceptable Timestamp Input", "[ERROR]: GroupBy validation failed. Please check that source has valid timestamps.")
+      assert(checkTs == "Acceptable Timestamp Input",
+             "[ERROR]: GroupBy validation failed. Please check that source has valid timestamps.")
     }
 
     val aggMetadata = if (groupByConf.aggregations != null) {
@@ -504,14 +504,16 @@ class Analyzer(tableUtils: TableUtils,
   def runTimestampCheck(df: DataFrame, sampleFraction: Double = 0.1): String = {
 
     // if timestamp column is present, sample if the timestamp values are not null
-    val checkTs = if ( df.schema.fieldNames.contains(Constants.TimeColumn) ) {
+    val checkTs = if (df.schema.fieldNames.contains(Constants.TimeColumn)) {
       val sumNotNulls = df
         .sample(sampleFraction)
         .agg(
           sum(when(col(Constants.TimeColumn).isNull, lit(0)).otherwise(lit(1))).cast(StringType).as("notNullCount")
         )
         .select(col("notNullCount"))
-        .rdd.collect()(0)(0).toString
+        .rdd
+        .collect()(0)(0)
+        .toString
       sumNotNulls
     } else {
       "No Ts Column"
@@ -519,7 +521,7 @@ class Analyzer(tableUtils: TableUtils,
 
     val tsStatus = checkTs match {
       case "0" => "Only Null Values" // this will trigger an assertion error
-      case _ => "Acceptable Timestamp Input"
+      case _   => "Acceptable Timestamp Input"
     }
 
     tsStatus
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -244,7 +244,7 @@ class Analyzer(tableUtils: TableUtils,
     }
 
     if (validationAssert) {
-      assert(checkTs == "Valid Timestamp Input", "ERROR: GroupBy validation failed. Please check that source has valid timestamps.")
+      assert(checkTs == "Acceptable Timestamp Input", "[ERROR]: GroupBy validation failed. Please check that source has valid timestamps.")
     }
 
     val aggMetadata = if (groupByConf.aggregations != null) {
@@ -517,12 +517,12 @@ class Analyzer(tableUtils: TableUtils,
       "No Ts Column"
     }
 
-    val tsStat = checkTs match {
-      case "0" => "Only Null Values"
-      case _ => "Valid Timestamp Input"
+    val tsStatus = checkTs match {
+      case "0" => "Only Null Values" // this will trigger an assertion error
+      case _ => "Acceptable Timestamp Input"
     }
 
-    tsStat
+    tsStatus
   }
 
   def run(): Unit =
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -194,7 +194,7 @@ class Analyzer(tableUtils: TableUtils,
     val name = "group_by/" + prefix + groupByConf.metaData.name
     logger.info(s"""|Running GroupBy analysis for $name ...""".stripMargin)
 
-    val checkTs = checkTimestamp(groupBy.inputDf)
+    val checkTs = runTimestampCheck(groupBy.inputDf)
 
     val analysis =
       if (enableHitter)
@@ -244,7 +244,7 @@ class Analyzer(tableUtils: TableUtils,
     }
 
     if (validationAssert) {
-      assert(checkTs == "Only Null Values", "ERROR: GroupBy validation failed. Please check that source has valid timestamps.")
+      assert(checkTs == "Valid Timestamp Input", "ERROR: GroupBy validation failed. Please check that source has valid timestamps.")
     }
 
     val aggMetadata = if (groupByConf.aggregations != null) {
@@ -501,7 +501,7 @@ class Analyzer(tableUtils: TableUtils,
 
   // For groupBys validate if the timestamp provided produces some values
   // if all values are null this should be flagged as an error
-  def checkTimestamp(df: DataFrame, sampleFraction: Double = 0.1): String = {
+  def runTimestampCheck(df: DataFrame, sampleFraction: Double = 0.1): String = {
 
     // if timestamp column is present, sample if the timestamp values are not null
     val checkTs = if ( df.schema.fieldNames.contains(Constants.TimeColumn) ) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: spark/src/main/scala/ai/chronon/spark/Analyzer.scala [hunk 1]
```diff
@@ -186,11 +186,15 @@ class Analyzer(tableUtils: TableUtils,
   def analyzeGroupBy(groupByConf: api.GroupBy,
                      prefix: String = "",
                      includeOutputTableName: Boolean = false,
-                     enableHitter: Boolean = false): (Array[AggregationMetadata], Map[String, DataType]) = {
+                     enableHitter: Boolean = false,
+                     validationAssert: Boolean = false): (Array[AggregationMetadata], Map[String, DataType]) = {
```
pengyu-hou, line 190: we want to perform sanity check for time column for all the cases. So we can remove this flag. 
As long as there is time column, then we should check it.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -238,6 +242,11 @@ class Analyzer(tableUtils: TableUtils,
            |""".stripMargin)
     }
 
+    if (validationAssert) {
+      assert(checkTs == "Acceptable Timestamp Input",
+             "[ERROR]: GroupBy validation failed. Please check that source has valid timestamps.")
+    }
```
pengyu-hou, line 248: As mentioned, we can remove the if condition to check it always.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -489,6 +499,34 @@ class Analyzer(tableUtils: TableUtils,
     }
   }
 
+  // For groupBys validate if the timestamp provided produces some values
+  // if all values are null this should be flagged as an error
+  def runTimestampCheck(df: DataFrame, sampleFraction: Double = 0.1): String = {
```
pengyu-hou, line 504: Could we also check if the timestamps are in milliseconds level? All the computation in the codebase is based on millisecond. 
Sometimes, users will use second or nanosecond, which will lead to null output values.

pengyu-hou (changes requested):
The flow looks good. Left a few comments. Please also add the time column sanity check for join. But it can come as a follow up PR.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pkundurthy: I can look into it. Though since the data-type is just Bigint, we can't be sure what the user actually intends to enter in general use-cases. If someone wants to use `1720000000` (2024-07-03 if in seconds ), is also valid millisec (1970-01-20), if using some historic data set. 

One thing that might be do-able is that we check if assumed milliseconds between some "reasonable timeframe" like:
Jan 1, 1970 00:00:00 UTC - Jan 1, 2100 00:00:00 UTC 
(assuming chronon will be around for ~100 years!)

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pkundurthy: Done.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pkundurthy: Done.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pkundurthy: @pengyu-hou - please take a look at a possible solution.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -489,6 +520,51 @@ class Analyzer(tableUtils: TableUtils,
     }
   }
 
+  // For groupBys validate if the timestamp provided produces some values
+  // if all values are null this should be flagged as an error
+  def runTimestampChecks(df: DataFrame, sampleFraction: Double = 0.1): Map[String, String] = {
```
pengyu-hou, line 525: 0.1 seems a large number. Let's change it to a smaller number with a upper bound if possible.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -191,6 +191,36 @@ class Analyzer(tableUtils: TableUtils,
...
+      )
+      assert(
+        timestampChecks("badRangeCount") == "0",
+        s"""[ERROR]: GroupBy validation failed.
+                 | Please check that source has valid epoch millisecond timestamps.
+                 | badRangeCount: ${timestampChecks("badRangeCount")}
+                 | """.stripMargin
+      )
+
+      logger.info(s"""ANALYSIS TIMESTAMP completed for group_by/${name}.
```
pengyu-hou, line 213: pretty cool.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -489,6 +520,51 @@ class Analyzer(tableUtils: TableUtils,
...
+      dataFrameToMap(
+        df.sample(sampleFraction)
+          .agg(
+            // will return 0 if all values are null
+            sum(when(col(Constants.TimeColumn).isNull, lit(0)).otherwise(lit(1)))
+              .cast(StringType)
+              .as("notNullCount"),
+            // assumes that we have valid unix milliseconds between the date range of
+            // 1970-01-01 00:00:00 (0L) to 2099-12-31 23:59:59 (4102473599999L)
+            // will return 0 if all values are within the range
```
pengyu-hou, line 538: Thanks for adding comments.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -489,6 +520,51 @@ class Analyzer(tableUtils: TableUtils,
...
+              .as("badRangeCount")
+          )
+          .select(col("notNullCount"), col("badRangeCount"))
+      )
+    } else {
+      Map(
+        "noTsColumn" -> "No Timestamp Column"
+      )
+    }
+
```
pengyu-hou, line 550:
```suggestion
```

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pengyu-hou: maybe we could do sample or first 1000 rows

pengyu-hou (commented):
The flow looks good overall. Left a few minor comments.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pengyu-hou: Somehow I missed this message. The approach works.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pkundurthy: done.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -489,6 +520,53 @@ class Analyzer(tableUtils: TableUtils,
...
+        df.limit(sampleN)
+          .agg(
+            // will return 0 if all values are null
+            sum(when(col(Constants.TimeColumn).isNull, lit(0)).otherwise(lit(1)))
+              .cast(StringType)
+              .as("notNullCount"),
+            // assumes that we have valid unix milliseconds between the date range of
+            // 1970-01-01 00:00:00 (0L) to 2099-12-31 23:59:59 (4102473599999L)
+            // will return 0 if all values are within the range
+            sum(when(col(Constants.TimeColumn).between(0L, 4102473599999L), lit(0)).otherwise(lit(1)))
```
hzding621, line 542: the current check wouldn't be able to detect milliseconds => seconds mistakes, because all seconds are also valid milliseconds epoch from ~1970s. But this is a very typical mistakes from users.

So I think in practice it might be better to add a lower bound cutoff at 31536000000L (which is milliseconds of 1971-01-01 00:00:00 UTC)

hzding621 (approved):

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pkundurthy: done.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -489,6 +520,53 @@ class Analyzer(tableUtils: TableUtils,
...
+      else { sampleNumber }
+      dataFrameToMap(
+        df.limit(sampleN)
+          .agg(
+            // will return 0 if all values are null
+            sum(when(col(Constants.TimeColumn).isNull, lit(0)).otherwise(lit(1)))
+              .cast(StringType)
+              .as("notNullCount"),
+            // assumes that we have valid unix milliseconds between the date range of
+            // 1971-01-01 00:00:00 (0L) to 2099-12-31 23:59:59 (4102473599999L)
```
hzding621, line 540: nit: 31536000000L (not 0L)

hzding621 (approved):

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pengyu-hou: good catch Haozhen.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pengyu-hou:
```suggestion
            // 1971-01-01 00:00:00 (31536000000L L) to 2099-12-31 23:59:59 (4102473599999L)
```

pengyu-hou (approved):
Great work here. 

Please address the last comment before merging: https://github.com/airbnb/chronon/pull/794#discussion_r1701169909

PR #795, opened 2024-07-23
fix: bug in new flow mode with bootstrap & derivation

hzding621:
## Summary

In join_part mode we overwrite join.left with bootstrap table before proceeding to the join_part backfill. However, bootstrap table and left table are NOT expected to have the same schema where there is an actual bootstrap_part in the join. The bootstrap table is expected to contain the joined result between left and various bootstrap_parts. 

The current code uses bootstrap_table's schema as left_schema when calculating BootstrapInfo, causing an issue in later logic. The new logic fixes it to restore to using left_table's schema as left_schema. 

We also add a safety check within BootstrapInfo.from for valid cases where left columns and base columns may overlap. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@donghanz @pengyu-hou @yuli-han

### DIFF
--- a/spark/src/test/scala/ai/chronon/spark/test/JoinFlowTest.scala
+++ b/spark/src/test/scala/ai/chronon/spark/test/JoinFlowTest.scala
@@ -49,7 +49,7 @@ class JoinFlowTest {
       .where(col("user").isNotNull)
       .save(transactionsTable)
 
-    val joinPart = Builders.JoinPart(groupBy = Builders.GroupBy(
+    val joinPart: JoinPart = Builders.JoinPart(groupBy = Builders.GroupBy(
       keyColumns = Seq("user"),
       sources = Seq(
         Builders.Source.events(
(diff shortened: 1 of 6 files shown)


nikhilsimha (approved):

nikhilsimha:
good fix! should we add a test case that would have failed?

hzding621:
@nikhilsimha thanks for review :) yeah this is integration-tested but i'm trying to create a UT for too

donghanz (approved):

pengyu-hou (approved):

spark/src/test/scala/ai/chronon/spark/test/JoinFlowTest.scala
```diff
@@ -0,0 +1,123 @@
...
+    assertTrue(tableUtils.tableExists(join.metaData.bootstrapTable))
+
+    // compute right
+    val joinPartJob = new Join(join, endDs, tableUtils, selectedJoinParts = Some(List(joinPart.fullPrefix)))
+    joinPartJob.computeJoinOpt(useBootstrapForLeft = true)
+    assertTrue(tableUtils.tableExists(join.partOutputTable(joinPart)))
+
+    // compute final
+    joinJob.computeFinal()
+    assertTrue(tableUtils.tableExists(join.metaData.outputTable))
```
hzding621, line 121: @donghanz FYI, I added a new UT that closely follows the Flow mode (left => join part => final). Previously this was not covered by UT

PR #799, opened 2024-07-29
refactor: join backfill refactor

donghanz:
## Summary

- remove tmp prefix in order to make the command more general. `/tmp/` prefix was used in Airflow. 
- add `extra_args` for extra arguments in run.py
- remove the dependency on adapters in JoinBackfill

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @yuli-han @pengyu-hou

### DIFF
--- a/api/py/ai/chronon/constants.py
+++ b/api/py/ai/chronon/constants.py
@@ -1,5 +1,0 @@
-from ai.chronon.scheduler.adapters.airflow_adapter import AirflowAdapter
-
-ADAPTERS = {
-    "airflow": AirflowAdapter,
-}
--- a/api/py/ai/chronon/repo/join_backfill.py
+++ b/api/py/ai/chronon/repo/join_backfill.py
@@ -86,7 +86,7 @@ class JoinBackfill:
     def command_template(self, config_path: str, extra_args: dict):
         if self.start_date:
             extra_args.update({"start_ds": self.start_date})
-        return f"""python3 /tmp/run.py --conf=/tmp/{config_path} --ds={self.end_date} \
+        return f"""run.py --conf={config_path} --ds={self.end_date} \
 {dict_to_bash_commands(extra_args)}"""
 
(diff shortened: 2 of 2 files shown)


hzding621 (approved):
👍

yuli-han (approved):
Hi @donghanz are we not having the adaptor concept here in join_backfill?

api/py/ai/chronon/constants.py
```diff
@@ -1,5 +0,0 @@
-from ai.chronon.scheduler.adapters.airflow_adapter import AirflowAdapter
```
pengyu-hou, line 1: question: 
does it mean we removed the dependency on Airflow?

pengyu-hou (approved):

donghanz:
> Hi @donghanz are we not having the adaptor concept here in join_backfill?

We extracted the adapter logic out of JoinBackfill to decouple these two. Now the adapter is handled before passing to the `JoinBackfill.run()` as an `orchestrator`. Otherwise, we have to import different adapters in the join backfill.

api/py/ai/chronon/constants.py
donghanz: We just decoupled `JoinBackfill` with `Airflow`. The `AirflowAdapter` or any other customized adapters can be passed through the [orchestrator](https://github.com/airbnb/chronon/pull/799/files#diff-70939bbcad59a3e132ccbbf2b5797512419bd014ea73bdd2ee1c6de879f2012dR124) argument in `JoinBackfill.run()`. This way, we don't have to import adapters in the `JoinBackfill` module.

pengyu-hou: Got it. 
yeah, this is better. Thanks @donghanz

PR #806, opened 2024-08-07
TTL cache retries more frequently on failures

yizzlez:
## Summary

This PR adds a new parameter to the TTL cache -- `failureTTLMillis`. This is a custom TTL for entries with type `Failure`. Currently `failureTTLMillis == ttlMillis`, which means this should have **no behavior change**. At Stripe, we will be changing some of the caches to have a significantly shorter `failureTTLMillis`.

## Why / Goal

At Stripe, we ran into an incident involving this particular piece of caching code for [groupByServing info](https://github.com/airbnb/chronon/blob/bd049312edb4e0f8a5a61711cfe5449b9d2e69fd/online/src/main/scala/ai/chronon/online/MetadataStore.scala#L168).

Our internal KV store was returning a handful of timeout errors for some requests. The issue is that this `Failure` is stored in a TTL Cache with a timeout of [2 hours](https://github.com/airbnb/chronon/blob/bd049312edb4e0f8a5a61711cfe5449b9d2e69fd/online/src/main/scala/ai/chronon/online/TTLCache.scala). This caused hosts to enter a bad state and error repeatedly, as it was fetching the previously stored error from the TTLCache instead of retrying against our KVStore. 

With this code change, it's possible to configure this TTL cache to have a `failureTTLMillis == 5s`, in which case we will automatically retry after 5s if we encounter a KV store timeout error.

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update


online/src/main/scala/ai/chronon/online/TTLCache.scala
```diff
@@ -33,7 +39,9 @@ class TTLCache[I, O](f: I => O,
                      contextBuilder: I => Metrics.Context,
                      ttlMillis: Long = 2 * 60 * 60 * 1000, // 2 hours
                      nowFunc: () => Long = { () => System.currentTimeMillis() },
-                     refreshIntervalMillis: Long = 8 * 1000 // 8 seconds
+                     refreshIntervalMillis: Long = 8 * 1000, // 8 seconds
+                     // same as ttlMillis, so behavior is unchanged barring an override
+                     failureTTLMillis: Long = 2 * 60 * 60 * 1000 // 2 hours
```
nikhilsimha, line 44: I think we should lower this to 30 seconds - I am no longer at airbnb, but I think it would benefit airbnb too.

online/src/main/scala/ai/chronon/online/TTLCache.scala
```diff
@@ -60,8 +68,14 @@ class TTLCache[I, O](f: I => O,
       contextBuilder(i).increment("cache.insert")
       entry.value
     } else {
+      val minFailureUpdateTTL = Math.min(intervalMillis, failureTTLMillis)
+      val shouldUpdate = entry.value match {
+        // Encountered a failure, update according to failure TTL.
+        case Failure(_) => nowFunc() - entry.updatedAtMillis > minFailureUpdateTTL
+        case _ => nowFunc() - entry.updatedAtMillis > intervalMillis
+      }
```
nikhilsimha, line 76:
```suggestion   
      val effectiveExpiry = entry.map(_ => intervalMillis).getOrElse(Math.min(intervalMillis, failureTTLMillis))       
```

online/src/main/scala/ai/chronon/online/TTLCache.scala
```diff
@@ -60,8 +68,14 @@ class TTLCache[I, O](f: I => O,
...
       if (
-        (nowFunc() - entry.updatedAtMillis > intervalMillis) &&
+        shouldUpdate &&
```
nikhilsimha, line 78:
```suggestion
        (nowFunc() - entry.updatedAtMillis > effectiveExpiry) &&
```

online/src/test/scala/ai/chronon/online/test/TTLCacheTest.scala
```diff
@@ -0,0 +1,138 @@
+package ai.chronon.online.test
```
nikhilsimha, line 1: Thanks a lot for adding this!

online/src/main/scala/ai/chronon/online/TTLCache.scala
nikhilsimha: minor simplification.

nikhilsimha (approved):
thanks for the change!

online/src/main/scala/ai/chronon/online/TTLCache.scala
hzding621: Hmm, I feel 30 seconds may be too frequent as default. Some of these TTLCache is used to cache metadata, which doesn't get updated frequently anyway. 

But I can see why for things like BatchIr cache, a more frequent failure refresh is desired.

pengyu-hou: Thanks @nikhilsimha. Our incident was caused by a stale but valid metadata. To mitigate it, we would have to flush the TTL cache. This should be addressed with @yuli-han 's recent work that we will only fetch active configs. 

I am curious what is the `failureTTLMillis` from Stripe side? @yizzlez 

For failure cases, I agree that we should use a lower TTL.

online/src/test/scala/ai/chronon/online/test/TTLCacheTest.scala
pengyu-hou: +1

PR #808, opened 2024-08-09
Update log for semantic hash

yuli-han:
## Summary

Update log for semantic hash. 
Use different log to distinguish the case when the table doesn't have semantic hash and when table has flag unset semantic hash.

Add table name in log, since we are computing semantic hash for join part table, final output table and left table in a single production job now. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@donghanz @hzding621 @pengyu-hou


hzding621 (approved):
@yuli-han thank you for fixing this!

donghanz (approved):

PR #811, opened 2024-08-14
[Backfill] Adding Archive Flag to TBLPROPERTIES

pkundurthy:
## Summary

- When a table is archived in spark, we add a new table property called `chronon_archived = true` 
- Explicitly setting the archive timestamp to UTC
- Add an option in the `alterTableProperties` to unset properties

## Why / Goal

- This allows us to flag all tables which are currently archived by Chronon
- Within airbnb, we can apply our own custom archiving rules based on this flag

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@donghanz @pengyu-hou

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
@@ -671,6 +671,9 @@ case class TableUtils(sparkSession: SparkSession) {
       val command = s"ALTER TABLE $tableName RENAME TO $finalArchiveTableName"
       logger.info(s"Archiving table with command: $command")
       sql(command)
+      val commandProp = s"ALTER TABLE $finalArchiveTableName SET TBLPROPERTIES ('chronon_archived' = 'true')"
+      logger.info(s"Update table properties to indicate archived table: $commandProp")
+      sql(commandProp)
     }
   }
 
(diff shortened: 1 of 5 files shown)


donghanz (approved):

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -382,6 +384,8 @@ case class TableUtils(sparkSession: SparkSession) {
     } else {
       if (tableProperties != null && tableProperties.nonEmpty) {
         alterTableProperties(tableName, tableProperties)
+        // remove any properties set if table was previously archived
+        unsetTableProperties(tableName, Seq(Constants.chrononArchiveFlag))
```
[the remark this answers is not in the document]
pkundurthy: changed this so that unsetting properties can be an optional input to alterTableProperties instead of a separate method.

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -47,7 +47,7 @@ case class TableUtils(sparkSession: SparkSession) {
   private val ARCHIVE_TIMESTAMP_FORMAT = "yyyyMMddHHmmss"
   @transient private lazy val archiveTimestampFormatter = DateTimeFormatter
     .ofPattern(ARCHIVE_TIMESTAMP_FORMAT)
-    .withZone(ZoneId.systemDefault())
+    .withZone(ZoneId.of("UTC"))
```
pkundurthy, line 50: forcing this to use UTC, since systemDefaults can be different between local, CI and production environments; we should be consistent.

donghanz (approved):

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
pkundurthy: @hzding621 - PTAL and let me know if you think this might cause issues.

PR #813, opened 2024-08-15
Analyzer - Adding skipTimestampCheck as an input option to the Analyzer

pkundurthy:
## Summary

 - We add a flag `skipTimestampCheck` to the Analyzer
 - We enable the Driver to be able to parse a --skip-timestamp-check boolean flag 
 - We reduce the default sample-size from 1000 ==> 100 for timestamp checks

## Why / Goal

 - We want to reduce the overhead introduced by timestamp checks
 - We want to provide an option to skip timestamp checks in the analyzer

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [x] Documentation update

### DIFF
--- a/docs/source/test_deploy_serve/Test.md
+++ b/docs/source/test_deploy_serve/Test.md
@@ -54,6 +54,8 @@ Optional parameters:
 
 `--enable-hitter`: enable skewed data analysis - include the heavy hitter analysis in output, only output schema if not specified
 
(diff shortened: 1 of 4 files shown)


pengyu-hou (commented):
it looks good. Please also make the sample limit to 100 rows.

pengyu-hou (approved):

yuli-han (approved):
Thanks for making the change!

hzding621 (approved):

PR #815, opened 2024-08-21
Disable timestamp check in metadata exporter

yuli-han:
## Summary

Disable timestamp check in metadata exporter. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@pkundurthy @hzding621

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -188,7 +188,7 @@ class Analyzer(tableUtils: TableUtils,
                      prefix: String = "",
                      includeOutputTableName: Boolean = false,
                      enableHitter: Boolean = false,
-                     skipTimestampCheck: Boolean = false): (Array[AggregationMetadata], Map[String, DataType]) = {
+                     skipTimestampCheck: Boolean = skipTimestampCheck): (Array[AggregationMetadata], Map[String, DataType]) = {
     groupByConf.setups.foreach(tableUtils.sql)
     val groupBy = GroupBy.from(groupByConf, range, tableUtils, computeDependency = enableHitter, finalize = true)
     val name = "group_by/" + prefix + groupByConf.metaData.name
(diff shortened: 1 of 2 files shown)


hzding621 (approved):

pkundurthy (approved):

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -184,11 +184,12 @@ class Analyzer(tableUtils: TableUtils,
...
-                     prefix: String = "",
-                     includeOutputTableName: Boolean = false,
-                     enableHitter: Boolean = false,
-                     skipTimestampCheck: Boolean = false): (Array[AggregationMetadata], Map[String, DataType]) = {
+  def analyzeGroupBy(
+      groupByConf: api.GroupBy,
+      prefix: String = "",
+      includeOutputTableName: Boolean = false,
+      enableHitter: Boolean = false,
+      skipTimestampCheck: Boolean = skipTimestampCheck): (Array[AggregationMetadata], Map[String, DataType]) = {
```
pkundurthy, line 192: @yuli-han - shouldn't this be Boolean = false?

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -258,11 +259,12 @@ class Analyzer(tableUtils: TableUtils,
...
-                  enableHitter: Boolean = false,
-                  validateTablePermission: Boolean = true,
-                  validationAssert: Boolean = false,
-                  skipTimestampCheck: Boolean = false): (Map[String, DataType], ListBuffer[AggregationMetadata]) = {
+  def analyzeJoin(
+      joinConf: api.Join,
+      enableHitter: Boolean = false,
+      validateTablePermission: Boolean = true,
+      validationAssert: Boolean = false,
+      skipTimestampCheck: Boolean = skipTimestampCheck): (Map[String, DataType], ListBuffer[AggregationMetadata]) = {
```
pkundurthy, line 267: same comment as above

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
yuli-han: Hi @pkundurthy I already set the skipTimestampCheck =  true when initialize the analyzer https://sourcegraph.a.musta.ch/github.com/airbnb/chronon/-/blob/spark/src/main/scala/ai/chronon/spark/MetadataExporter.scala?L53 If we set default value to be false we have to pass the skipTimestampCheck again when calling the analyzeJoin function. It should be equal to the skipTimestampCheck default value when we initialize the analyzer right?

PR #817, opened 2024-08-27
fix: join keys filtering

donghanz:
## Summary

There are a few issues on the join keys filtering:
1. Bloom Filter
- We only have a unified join level bloom filter. However, now we compute Join Part tasks separately, so it makes more sense to generate the bloom filter per JP.
- bootstrap is always enabled once `useCachedLeft` is true, even though there is no actual bootstrap tables defined. Hence, we still need to generate the bloom filter, even though there is `matched_hashes` column.
2. SQL injection in the small mode
- whether the job should run in the small mode or not depends on the actual row count that needs to be backfilled. Especially in the case of bootstrap, the actual row count may far less than the left table row count. Hence, we move the small mode logic to the post-bootstrap step.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested: Tested with internal datasets

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @pengyu-hou @yuli-han

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Join.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Join.scala
@@ -340,11 +340,18 @@ class Join(joinConf: api.Join,
                     s"Macro ${Constants.ChrononRunDs} is only supported for single day join, current range is ${leftRange}")
                 }
 
-                val (runSmallMode: Boolean, bloomFilterOpt: Option[util.Map[String, BloomFilter]], filteredJoinPart: api.JoinPart) =
+                val (runSmallMode: Boolean,
+                     bloomFilterOpt: Option[util.Map[String, BloomFilter]],
+                     filteredJoinPart: api.JoinPart) =
                   genKeyFilter(joinPart, unfilledLeftDf)
 
                 val df =
-                  computeRightTable(unfilledLeftDf, filteredJoinPart, leftRange, leftTimeRangeOpt, bloomFilterOpt, runSmallMode)
+                  computeRightTable(unfilledLeftDf,
+                                    filteredJoinPart,
+                                    leftRange,
+                                    leftTimeRangeOpt,
+                                    bloomFilterOpt,
+                                    runSmallMode)
                     .map(df => filteredJoinPart -> df)
                 Thread.currentThread().setName(s"done-$threadName")
                 df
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -239,7 +236,7 @@ abstract class JoinBase(joinConf: api.Join,
     val rightBloomMap = if (skipBloom) {
       None
     } else {
-      JoinUtils.genBloomFilterIfNeeded(joinPart, joinConf, rowCount, unfilledRange, joinLevelBloomMapOpt)
+      JoinUtils.genBloomFilterIfNeeded(joinPart, joinConf, rowCount, unfilledRange, bloomMapOpt)
```
hzding621, line 239: nit: I think this function no longer generate BF but simply pull BFs for the corresponding key, so we can rename it

spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -188,14 +188,11 @@ abstract class JoinBase(joinConf: api.Join,
             val leftUnfilledRange = unfilledRange.shift(-shiftDays)
             val prunedLeft = leftDf.flatMap(_.prunePartitions(leftUnfilledRange))
             val filledDf =
-              computeJoinPart(prunedLeft, joinPart, joinLevelBloomMapOpt, smallMode)
+              computeJoinPart(prunedLeft, joinPart, bloomMapOpt, smallMode)
```
hzding621, line 191: nit: I don't think we need to pass in smallMode anymore? smallMode and bloomMapOpt are always mutually exclusive

spark/src/main/scala/ai/chronon/spark/Join.scala
```diff
@@ -317,13 +300,41 @@ class Join(joinConf: api.Join,
...
+                  } else {
+                    false
+                  }
+                }
+
+                // create filter for leftDf
                 val bloomFilterOpt = if (runSmallMode) {
                   // If left DF is small, hardcode the key filter into the joinPart's GroupBy's where clause.
-                  injectKeyFilter(leftDf, joinPart)
+                  injectKeyFilter(unfilledLeftDf.map(_.df).get, joinPart)
```
hzding621, line 322: nit: can we move this line 322 to after line 308? 

`injectKeyFilter` is a function with side effect (it modifies joinPart in place) so it's risky and should be made obvious. right now it's part of computing `bloomFilterOpt` which is confusing. Let's refactor this part.

hzding621 (approved):
🎉 great fix!!!

yuli-han (approved):

PR #818, opened 2024-08-27
Add deprecation date in metadata 

yuli-han:
## Summary

Add an optional string field deprecationDate in metadata. Users can set up the deprecationDate in their python config and the data will be saved in metadata in json config. 

The optiona string filed has to be in format like YYYY-mm-DD, add validation code in validator to make sure the format is correct. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @pengyu-hou

### DIFF
--- a/api/py/ai/chronon/group_by.py
+++ b/api/py/ai/chronon/group_by.py
@@ -367,6 +367,7 @@ def GroupBy(
     name: str = None,
     tags: Dict[str, str] = None,
     derivations: List[ttypes.Derivation] = None,
+    deprecationDate: str = None,
     **kwargs,
 ) -> ttypes.GroupBy:
     """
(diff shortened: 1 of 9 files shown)


api/thrift/api.thrift
```diff
@@ -274,7 +276,7 @@ struct GroupBy {
     5: optional Accuracy accuracy
     // Optional start date for a group by backfill, if it's unset then no historical partitions will be generate
     6: optional string backfillStartDate
-    // support for offline only for now
```
yuli-han, line 277: Removed since derivation is support for online groupBy now

api/py/ai/chronon/group_by.py
```diff
@@ -367,6 +367,7 @@ def GroupBy(
     name: str = None,
     tags: Dict[str, str] = None,
     derivations: List[ttypes.Derivation] = None,
+    deprecationDate: str = None,
```
hzding621, line 370: snake case

hzding621:
@yuli-han Can you add more context on why deprecationDate should be a defined field in metadata instead of part of custom_json? i.e. Are we going to do something about the deprecation_date, e.g. warning or error when compiling against a config that is soon to be deprecated or already deprecated?

hzding621 (approved):

donghanz (approved):

PR #823, opened 2024-09-03
fix: improve table archive in jp parallel mode 

hzding621:
## Summary

In JP parallel mode, in some edge case there is a bug that will cause unwanted archives. 

When the left hash changes:
- the `left` task will handle archive of the bootstrap table
- the `right` and `final` task should not do the archive again! But because `right` and `final` tasks are using semantic_hash stored on the final table (which is not updated), it will do the archive again. This will cause the bootstrap_table to be archived again after it's already re-computed by the `left` task, and this will result in data loss. 

Besides the fix, this PR also adds more debugging logs. 
## Why / Goal

Tables are being incorrectedly archived right now. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@donghanz


donghanz (approved):
Thanks for the fix!

PR #824, opened 2024-09-06
fix: join derivations on structs

hzding621:
## Summary

Fix catalyst util handling struct inputs. Currently all structs are in Array format after converting from Avro format from KVStore: https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Row.scala#L56-L68

## Why / Goal

Fix derivations with struct inputs 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@yuli-han

### DIFF
--- a/online/src/main/scala/ai/chronon/online/SparkInternalRowConversions.scala
+++ b/online/src/main/scala/ai/chronon/online/SparkInternalRowConversions.scala
@@ -116,7 +116,7 @@ object SparkInternalRowConversions {
   // The purpose of this class is to be used on fetcher output in a fetching context
   // we take a data type and build a function that operates on actual value
   // we want to build functions where we only branch at construction time, but not at function execution time.
-  def to(dataType: types.DataType, structToMap: Boolean = true): Any => Any = {
+  def to(dataType: types.DataType, structToMap: Boolean): Any => Any = {
     val unguardedFunc: Any => Any = dataType match {
       case types.MapType(keyType, valueType, _) =>
         val keyConverter = to(keyType, structToMap)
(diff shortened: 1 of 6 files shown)


yuli-han (approved):
Thanks for the hot fix!

online/src/main/scala/ai/chronon/online/SparkInternalRowConversions.scala
```diff
@@ -116,7 +116,7 @@ object SparkInternalRowConversions {
   // The purpose of this class is to be used on fetcher output in a fetching context
   // we take a data type and build a function that operates on actual value
   // we want to build functions where we only branch at construction time, but not at function execution time.
-  def to(dataType: types.DataType, structToMap: Boolean = true): Any => Any = {
+  def to(dataType: types.DataType, structToMap: Boolean): Any => Any = {
```
yuli-han, line 119: do we need to mark the internal row as structToMap = false?

hzding621: @yuli-han WDYM?

hzding621: so I ended up not hard coding structToMap in this function, but passing the structToMap value down. I think that's more intuitive. 

Instead, the outside caller of `to` will specify structToMap, so:
- inputEncoder will set it to true
- inputArrEncoder will set it to false

donghanz (approved):

PR #825, opened 2024-09-06
Update the online derivation validation behavior

yuli-han:
## Summary

We used to exclude the keys columns for online joins/group_bys because keys are not included in the online fetching result. But there are some use cases where the user firstly add keys in offline group_bys derivations, then mark the offline group_by to be online. The validator failed in this case because we are not allowed online derivation on keys. We don't want users to change their offline derivation behavior just to pass the validator and recompute everything. 

This PR is to update the validator behavior to enable online derivations on keys. Although online derivation on keys are not gonna make any difference in fetching result, it won't break the re-compiling on existing offline entity. 

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @donghanz

### DIFF
--- a/api/py/ai/chronon/repo/validator.py
+++ b/api/py/ai/chronon/repo/validator.py
@@ -347,12 +347,8 @@ class ChrononRepoValidator(object):
         # Only validate the join derivation when the underlying groupBy is valid
         if join.derivations and group_by_correct:
             features = get_pre_derived_join_features(join)
-            # For online joins keys are not included in output schema
-            if join.metaData.online:
-                columns = features
-            else:
-                keys = get_pre_derived_source_keys(join.left)
-                columns = features + keys
+            keys = get_pre_derived_source_keys(join.left)
+            columns = features + keys
             errors.extend(self._validate_derivations(columns, join.derivations))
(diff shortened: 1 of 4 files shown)


hzding621:
@yuli-han it seems we do allow online derivations on keys after all? https://github.com/airbnb/chronon/blob/main/online/src/main/scala/ai/chronon/online/OnlineDerivationUtil.scala#L43

otherwise, user's gb in the scenario you described would fail online, right?

yuli-han:
@hzding621 Yep! By disable I mean currently the compiler will fail if users change their offline group_by with derivation on keys to online. But the online derivation on keys is always supported from the fetcher side. So I make this change to enable this in compiler as well.

yuli-han:
Confirmed that online derivation supports derivation on keys

pkundurthy:
Do we need to update any documentation to remove any previous notes on validation failing on offline -> online change? @yuli-han 
LGTM: stamped.

pkundurthy (approved):

PR #827, opened 2024-09-11
fix: fix backfill end ds when endPartition is defined

donghanz:
## Summary

If `leftSource.query.endPartition` is later than `endPartition`, then `endPartition` will be ignored. This results in the backfill range is larger than expected.

To resolve this, we select the earlier date as the backfill end date.

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
@@ -111,7 +111,7 @@ object JoinUtils {
     lazy val defaultLeftStart = Option(leftSource.query.startPartition)
       .getOrElse(tableUtils.firstAvailablePartition(leftSource.table, leftSource.subPartitionFilters).get)
     val leftStart = overrideStart.getOrElse(defaultLeftStart)
-    val leftEnd = Option(leftSource.query.endPartition).getOrElse(endPartition)
+    val leftEnd = Seq(Option(leftSource.query.endPartition), Some(endPartition)).flatten.min
     PartitionRange(leftStart, leftEnd)(tableUtils)
   }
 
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
```diff
@@ -111,7 +111,7 @@ object JoinUtils {
     lazy val defaultLeftStart = Option(leftSource.query.startPartition)
       .getOrElse(tableUtils.firstAvailablePartition(leftSource.table, leftSource.subPartitionFilters).get)
     val leftStart = overrideStart.getOrElse(defaultLeftStart)
-    val leftEnd = Option(leftSource.query.endPartition).getOrElse(endPartition)
+    val leftEnd = Seq(Option(leftSource.query.endPartition), Some(endPartition)).flatten.min
```
rohitgirme, line 114: Why is `endPartition` treated differently from `overrideStartPartition`? Seems like the intention of both is to override the dates from leftSource? Seems a bit confusing to first timers

pengyu-hou (approved):

spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
pengyu-hou: For the right boundary of the backfill range, we still want to use end partition even if it is earlier than the query.endPartition.

yuli-han (approved):

spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
donghanz: `leftSource.query.endPartition` has a different purpose compared with `overrideStartPartition`. The former indicates the latest valid ds partition of the table (after which the table may have no data); the latter is designed to override the backfill range.

donghanz: Having a earlier `leftEnd` does not violate the configuration - `leftSource.query.endPartition`

PR #832, opened 2024-09-19
feat: support repartition for chaining streaming job

hzding621:
## Summary

Add a prop to support repartition (to a higher parallelism) in chaining job. 

## Why / Goal

Chaining job uses micro batching and thus support repartition to increase parallelism. This will help reduce lag. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

Tested on Airbnb side on user chaining job and observed improved latency/lag after increasing the parallelism 

## Checklist
- [ ] Documentation update

## Reviewers

@pengyu-hou @yuli-han

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/streaming/JoinSourceRunner.scala
+++ b/spark/src/main/scala/ai/chronon/spark/streaming/JoinSourceRunner.scala
@@ -16,7 +16,6 @@
 
 package ai.chronon.spark.streaming
 
-import org.slf4j.LoggerFactory
 import ai.chronon.api
 import ai.chronon.api.Extensions.{GroupByOps, SourceOps}
 import ai.chronon.api._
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/streaming/JoinSourceRunner.scala
```diff
@@ -284,13 +287,23 @@ class JoinSourceRunner(groupByConf: api.GroupBy, conf: Map[String, String] = Map
     Some(sorted(k))
   }
 
+  def repartition(stream: DataStream): DataStream = {
+    if (microBatchRepartition <= 0) {
+      stream
+    } else {
+      logger.info(s"Repartitioning stream to $microBatchRepartition partitions")
```
pengyu-hou, line 294: nit: maybe use debug log level?
```suggestion
      logger.debug(s"Repartitioning stream to $microBatchRepartition partitions")
```

hzding621: hmm well maybe info is still better. debug i feel is for extra granular, redundant information that we don't want to show in production.

pkundurthy (approved):

PR #848, opened 2024-10-01
Use builder pattern to create java fetcher and add a default log publish error rate of 0.1%

pengyu-hou:
## Summary

We noticed log publish failures but the exception message is not clear. 
This PR will sample the error logs at 0.1% by default is no sample percentage is set in the metadata. Otherwise, it will take the number from the metadata.samplePercent.

To simplify the java fetcher debug process, refactored the code to use builder pattern so that different params can be set separately. 

## Why / Goal

Improve debuggability 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@airbnb/zipline-maintainers

### DIFF
--- a/online/src/main/java/ai/chronon/online/JavaFetcher.java
+++ b/online/src/main/java/ai/chronon/online/JavaFetcher.java
@@ -46,6 +46,57 @@ public class JavaFetcher {
     this.fetcher = new Fetcher(kvStore, metaDataSet, timeoutMillis, logFunc, false, registry, callerName, flagStore, disableErrorThrows);
   }
 
+  // user builder pattern to create JavaFetcher
+  private JavaFetcher(Builder builder) {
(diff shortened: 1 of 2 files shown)


hzding621 (approved):

online/src/main/java/ai/chronon/online/JavaFetcher.java
```diff
@@ -32,130 +32,196 @@
...
+
+    /* user builder pattern to create JavaFetcher
+    example way to create the java fetcher
+    JavaFetcher fetcher = new JavaFetcher.Builder(kvStore, metaDataSet, timeoutMillis, logFunc, registry)
+                                        .callerName(callerName)
+                                        .flagStore(flagStore)
+                                        .disableErrorThrows(disableErrorThrows)
+                                        .build();
+     */
+    private JavaFetcher(Builder builder) {
```
pengyu-hou, line 57: it was due to IDE reformat. This is the gist of the builder change in java fetcher.

PR #853, opened 2024-10-02
[Backfill]Remove exception when the keys are all null in unfilledLeftDf

yuli-han:
## Summary

When we run the join part backfill on the left, we used to error out when the key are all empty in left table. This doesn't work when we run the join part backfill after bootstrap, because the unfilledLeftDf keys can be all null. 

This PR is to remove the exception throw statement in genKeyFilter function. When the keys in unfilledLeftDf are all empty, we will build an empty filter(`where keys in ()`) in join part and the computeRightTable function will return an empty dataframe, like the case when the bootstrap table covers everything. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @donghanz @pengyu-hou

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
@@ -304,11 +304,10 @@ object JoinUtils {
               s"KeyName: $keyName, leftSide KeyName: $leftSideKeyName , Join right to left: ${joinPart.rightToLeft
                 .mkString(", ")}")
             val values = collectedLeft.map(row => row.getAs[Any](leftSideKeyName))
-            // Check for null keys, warn if found, err if all null
+            // Check for null keys, warn if found
             val (notNullValues, nullValues) = values.partition(_ != null)
             if (notNullValues.isEmpty) {
-              throw new RuntimeException(
-                s"No not-null keys found for key: $keyName. Check source table or where clauses.")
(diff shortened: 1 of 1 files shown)


hzding621:
@yuli-han just tried it in spark sql cli, I think this syntax will fail in spark 

```
where keys in ()
```

spark/src/main/scala/ai/chronon/spark/JoinUtils.scala
```diff
@@ -319,8 +318,11 @@ object JoinUtils {
               case other     => other.toString // Keep other types (like Int) as they are
             }.toSet
 
-            // Form the final WHERE clause for injection
-            s"$groupByKeyExpression in (${valueSet.mkString(sep = ",")})"
+            if (valueSet.isEmpty) {
```
hzding621, line 321: `valueSet.nonEmpty`

yuli-han: @hzding621 this should be isEmpty right? Because the value set will be empty if the notNullValues is empty and in this case we should add the filter as false?

hzding621: yeah this looks right now. I think I commented before seeing your commit https://github.com/airbnb/chronon/pull/853/commits/2f51a6208049b8889506d5332e242f637f544931

hzding621 (approved):

PR #855, opened 2024-10-04
fix: Allow upload job to complete when output data is empty

hzding621:
## Summary

Allow GroupByUpload job to complete when output data is empty except for the `group_by_serving_info` row. 

## Why / Goal

This will ensure that `group_by_serving_info` is being updated into KV store such that metadata is kept up to date. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Reviewers

@pengyu-hou @donghanz

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/GroupByUpload.scala
+++ b/spark/src/main/scala/ai/chronon/spark/GroupByUpload.scala
@@ -242,9 +242,16 @@ object GroupByUpload {
 
     val metricRow =
       kvDfReloaded.selectExpr("sum(bit_length(key_bytes))/8", "sum(bit_length(value_bytes))/8", "count(*)").collect()
-    context.gauge(Metrics.Name.KeyBytes, metricRow(0).getDouble(0).toLong)
-    context.gauge(Metrics.Name.ValueBytes, metricRow(0).getDouble(1).toLong)
-    context.gauge(Metrics.Name.RowCount, metricRow(0).getLong(2))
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/ai/chronon/spark/GroupByUpload.scala
```diff
@@ -242,9 +242,17 @@ object GroupByUpload {
...
+
+    val rowCount = metricRow(0).getLong(2)
+    context.gauge(Metrics.Name.RowCount, rowCount)
+    // Allow for upload data to be empty so that we can still update the GroupByServingInfo
+    // TODO: in the long run, consider separating GroupByServingInfo into a separate job
+    if (rowCount > 0) {
+      context.gauge(Metrics.Name.KeyBytes, metricRow(0).getDouble(0).toLong)
+      context.gauge(Metrics.Name.ValueBytes, metricRow(0).getDouble(1).toLong)
+    } else {
+      logger.warn(s"Empty upload for ${groupByConf.metaData.name} at $endDs")
```
rohitgirme, line 254: Assuming fetcher has no assumptions on data being present?

hzding621: @rohitgirme Yeah fetcher won't fail, it will just not be able to fetch any data which is the expected behavior.

donghanz (approved):

PR #857, opened 2024-10-04
fix: make logging codec thread safe

hzding621:
## Summary

`AvroCodec` is not thread-safe, and in particular, the `outputStream = new ByteArrayOutputStream()` is a point of contention. Currently we use `keyCodec` and `valueCodec` which are cached in `getJoinCodecs` and accessed across multiple threads during the logging stage. During logging, we work-around errors by doing brute-force encode and try-decode, up to 3 trials, to ensure the encoding process succeeds, which was definitely not an ideal solution ([code](https://github.com/airbnb/chronon/blob/v0.0.83/online/src/main/scala/ai/chronon/online/Fetcher.scala#L300)). 

The fix is to move the construction (the `AvroCodec.of` call) inside logging, which will generate thread local cached instances of codecs. 

## Why / Goal

Attempt to fix a production logging failure issue, which only affects a particular join. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@yuli-han @pengyu-hou @donghanz @rohitgirme

### DIFF
--- a/online/src/main/scala/ai/chronon/online/Fetcher.scala
+++ b/online/src/main/scala/ai/chronon/online/Fetcher.scala
@@ -159,10 +159,8 @@ class Fetcher(val kvStore: KVStore,
 
     val joinName = joinConf.metaData.nameToFilePath
     val keySchema = StructType(s"${joinName.sanitize}_key", keyFields.toArray)
-    val keyCodec = AvroCodec.of(AvroConversions.fromChrononSchema(keySchema).toString)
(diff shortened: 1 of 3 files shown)


donghanz (approved):

pengyu-hou (approved):
LGTM. Thanks for picking this up.

PR #863, opened 2024-10-15
fix: skip nulls in the right table

donghanz:
## Summary

fix: skip nulls in the right table

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Join.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Join.scala
@@ -212,9 +212,19 @@ class Join(joinConf: api.Join,
       val wheres = Seq(s"ds >= '${effectiveRange.start}'", s"ds <= '${effectiveRange.end}'")
       val sql = QueryUtils.build(null, partTable, wheres)
       logger.info(s"Pulling data from joinPart table with: $sql")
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/Join.scala
```diff
@@ -212,9 +212,19 @@ class Join(joinConf: api.Join,
...
-      val df = tableUtils.sparkSession.sql(sql)
-      (joinPart, df)
-    }
+      val dfTry = Try {
+        tableUtils.sparkSession.sql(sql)
+      }
+
+      dfTry match {
+        case Success(df) => Some((joinPart, df))
+        case Failure(e: org.apache.spark.sql.AnalysisException) if e.getMessage.contains("Table or view not found") =>
```
hzding621, line 221: instead of relying on string matching, can we explicitly check existence by using `TableUtils.tableExists`?

hzding621: Pushed the change

hzding621 (approved):

yuli-han (approved):

PR #868, opened 2024-10-18
Join failed to get groupBy module on different file structure

MickJermsurawong:
## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/py/ai/chronon/join.py
+++ b/api/py/ai/chronon/join.py
@@ -63,7 +63,7 @@ def JoinPart(
     gc.collect()
     group_by_module_name = None
     for ref in gc.get_referrers(group_by):
-        if "__name__" in ref and ref["__name__"].startswith("group_bys"):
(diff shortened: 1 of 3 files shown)


nikhilsimha:
ah, the fix here is to set python path to your top level chronon configuration directory.

if vscode recognizes the path, compile.py should too.

MickJermsurawong:
thanks nikhil, yup appending python path works.. for now i have to add add pythonpath in unit test, wrapping for compile.py, and add ignore on mypy. I think it would be great if we can treat these definitions like native python resources using the standard python module references

nikhilsimha:
They should work like standard python references actually - it is just that the python env needs to be setup to recognize chronon on the module. Or maybe I am misunderstanding, is it a wrong assumption for the framework to expect the top level chronon directory (`sample` in this case) to be on python path?

PR #869, opened 2024-10-21
Add support for Delta lake tables in TableUtils

piyush-zlai:
## Summary

Add delta lake support in Chronon. This PR refactors TableUtils a bit to allow us to support Delta Lake (and other table formats in the future like Hudi). The crux of the PR is the refactor of TableUtils - I've pulled out some of the format specific stuff into per-format objects (e.g. Hive / Iceberg / Delta) to help us cover aspects like creating tables, listing partitions. 

As the versions of the various libraries we build against in Chronon is on the older side (e.g. delta lake - 2.0.x for Spark 3.2), we've structured the code to allow users to override their format providers. This allows folks say on Delta 3.x (Spark 3.5) to write their own Delta3xFormatProvider that is built against the newer version and configure it via the `spark.chronon.table.format_provider` Spark config setting. For users on the same versions / rails as OSS, they can skip configuring this and we'll default to the DefaultFormatProvider (which maintains the existing Iceberg / Hive behavior along with support for delta).  

Notes on testing:
* TableUtilsTest is un-changed. This test covers a bunch of Hive related tests (like creating tables, inserting partitions) and some format agnostic stuff (like schema evolution type of checks). 
* Added a TableUtilsFormat test - for now this covers testing Delta. Can add Iceberg in a follow up (as Iceberg format support is currently untested in the project). 
* Unfortunately the way we create Spark sessions in not ideal in the Chronon project. We kick off *all* our tests in parallel at the same time in CI (same JVM) and we end up triggering code like: [SchemaEvolutionTest](https://github.com/airbnb/chronon/blob/main/spark/src/test/scala/ai/chronon/spark/test/SchemaEvolutionTest.scala#L65) where we are creating the SparkSession at the class object construction time. This results in many tests vying to create sessions at the same time - most with the same spark settings but some like the new delta format test with different settings (using the delta catalog, delta spark session extension). Thanks to this we're not able to run the delta tests in the test JVM as the others AND we need to ensure that the other test spark sessions don't end up getting created with the wrong settings as that will cause the delta test to fail. 
* The potentially nice way to handle this would be to refactor all our existing spark tests to postpone session creation till the test body and handle session cleanup etc but as this is a fairly involved / risky change, I've chosen to instead use an env var to have the Spark session builder choose the right delta / non-delta options. It's not the cleanest but it does minimize the blast radius of these changes considerably. 

## Why / Goal

Allow folks keen on experimenting with Chronon that use delta lake to do so. Set the stage for supporting additional table formats in the future.

## Test Plan

- [X] Added Unit Tests
- [X] Covered by existing CI
- [ ] Integration tested

@mickjermsurawong-openai was also able to pull these changes and test things out using delta lake tables. We were able to sort out issues with Delta 3.x (as they're on Spark 3.5) using the provider interface. 

## Checklist
- [ ] Documentation update

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
                       conda activate chronon_py
                       # Increase if we see OOM.
                       export SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=4G -Xmx4G -Xms2G"
-                      sbt "++ 2.12.12 test"
+                      sbt "++ 2.12.12 'test -- -l ai.chronon.spark.test.TableUtilsFormatTest'"
             - store_test_results:
                   path: /chronon/spark/target/test-reports
             - store_test_results:
@@ -63,7 +63,7 @@ jobs:
                       conda activate chronon_py
                       # Increase if we see OOM.
                       export SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=4G -Xmx4G -Xms2G"
-                      sbt "++ 2.13.6 test"
+                      sbt "++ 2.13.6 'test -- -l ai.chronon.spark.test.TableUtilsFormatTest'"
             - store_test_results:
                   path: /chronon/spark/target/test-reports
             - store_test_results:
@@ -77,6 +77,32 @@ jobs:
                   path: /tmp/spark-warehouse.tar.gz
                   destination: spark_warehouse.tar.gz
                   when: on_fail
+
+    "Scala 13 -- Table Format Tests":
+      executor: docker_baseimg_executor
+      steps:
+        - checkout
+        - run:
+            name: Run Scala 13 table format tests
+            shell: /bin/bash -leuxo pipefail
+            command: |
+              conda activate chronon_py
+              # Increase if we see OOM.
+              export SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=4G -Xmx4G -Xms2G"
+              sbt "++ 2.13.6 'testOnly ai.chronon.spark.test.TableUtilsFormatTest'"
+        - store_test_results:
+            path: /chronon/spark/target/test-reports
+        - store_test_results:
+            path: /chronon/aggregator/target/test-reports
+        - run:
+            name: Compress spark-warehouse
+            command: |
+              cd /tmp/ && tar -czvf spark-warehouse.tar.gz chronon/spark-warehouse
+            when: on_fail
+        - store_artifacts:
+            path: /tmp/spark-warehouse.tar.gz
+            destination: spark_warehouse.tar.gz
+            when: on_fail
 
     "Scala 11 -- Compile":
       executor: docker_baseimg_executor
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
                       conda activate chronon_py
                       # Increase if we see OOM.
                       export SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=4G -Xmx4G -Xms2G"
-                      sbt "++ 2.12.12 'test -- -l ai.chronon.spark.test.TableUtilsFormatTest'"
+                      sbt "++ 2.12.12 test"
             - store_test_results:
                   path: /chronon/spark/target/test-reports
             - store_test_results:
@@ -63,7 +63,7 @@ jobs:
                       conda activate chronon_py
                       # Increase if we see OOM.
                       export SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=4G -Xmx4G -Xms2G"
-                      sbt "++ 2.13.6 'test -- -l ai.chronon.spark.test.TableUtilsFormatTest'"
+                      sbt "++ 2.13.6 test"
             - store_test_results:
                   path: /chronon/spark/target/test-reports
             - store_test_results:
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -41,6 +41,76 @@ import scala.collection.immutable
...
+
+  override def partitions(tableName: String)(implicit sparkSession: SparkSession): Seq[Map[String, String]] = {
+    // data is structured as a Df with single composite partition key column. Every row is a partition with the
+    // column values filled out as a formatted key=value pair
+    // Eg. df schema = (partitions: String)
+    // rows = [ "day=2020-10-10/hour=00", ... ]
+    sparkSession.sqlContext
+      .sql(s"SHOW PARTITIONS $tableName")
+      .collect()
+      .map(row => parseHivePartition(row.getString(0)))
```
mickjermsurawong-openai, line 78: Curious if the partition keys here are enforced by Chronon? i.e I'm wondering if users can specified partition key and we have increasing cardinality of this and we end up returning large entries of this

piyush-zlai: the partition keys chosen / used here are based on the `partitionColumn` that you configure in the table util setup for your job(s). You are right that you could have a very large number of partitions here loaded up in mem.

.circleci/config.yml
```diff
@@ -78,6 +78,35 @@ jobs:
                   destination: spark_warehouse.tar.gz
                   when: on_fail
 
+    # run these separately as we need a isolated JVM to not have Spark session settings interfere with other runs
```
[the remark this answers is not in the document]
piyush-zlai: done

spark/src/main/scala/ai/chronon/spark/SparkSessionBuilder.scala
```diff
@@ -38,6 +38,24 @@ object SparkSessionBuilder {
             localWarehouseLocation: Option[String] = None,
             additionalConfig: Option[Map[String, String]] = None,
             enforceKryoSerializer: Boolean = true): SparkSession = {
+
+    // allow us to override the format by specifying env vars. This allows us to not have to worry about interference
+    // between Spark sessions created in existing chronon tests that need the hive format and some specific tests
+    // that require a format override like delta lake.
+    val (formatConfigs, kryoRegistrator) = sys.env.get(FormatTestEnvVar) match {
```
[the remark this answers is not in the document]
piyush-zlai: We could - I actually thought the code is cleaner with the match up front - this allows us to skip the use of vars and a lot of if/else type of checks in a couple of places. 
Let me know if you feel strongly and I can add there.

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -41,6 +43,165 @@ import scala.collection.immutable
...
+  private def isDeltaTable(tableName: String): Boolean = {
+    Try {
+      val describeResult = sparkSession.sql(s"DESCRIBE DETAIL $tableName")
+      describeResult.select("format").first().getString(0).toLowerCase
+    } match {
+      case Success(format) =>
+        logger.info(s"Delta check: Successfully read the format of table: $tableName as $format")
+        format == "delta"
+      case _ =>
+        // the describe detail calls fails for Iceberg tables
```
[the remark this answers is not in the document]
piyush-zlai: ah yeah, nice catch

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -134,67 +294,46 @@ case class TableUtils(sparkSession: SparkSession) {
...
       }
+    }
   }
 
   def partitions(tableName: String, subPartitionsFilter: Map[String, String] = Map.empty): Seq[String] = {
     if (!tableExists(tableName)) return Seq.empty[String]
-    if (isIcebergTable(tableName)) {
+    val format = tableReadFormat(tableName)
+
+    if (format == Iceberg) {
```
mears-stripe, line 317: Consider having a supportSubPartitionsFilter method in Format. Stripe has another format that we would have (temp views) that also would not support subPartitionFilters.

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -134,67 +294,46 @@ case class TableUtils(sparkSession: SparkSession) {
...
-            subPartitionsFilter.forall {
-              case (k, v) => partitionMap.get(k).contains(v)
-            }
-          ) {
-            partitionMap.get(partitionColumn)
-          } else {
-            None
-          }
+
+    val partitionSeq = format.partitions(tableName)(sparkSession)
```
mears-stripe, line 324: This logic could live in `Format` in a method with a concrete implementation with this as the default. The method could then be overridden for `Formats` like Iceberg that need different logic (`getIcebergPartitions` for Iceberg).

mears-stripe (changes requested):

mears-stripe (commented):
Excited for this change!

spark/src/test/scala/ai/chronon/spark/test/TableUtilsFormatTest.scala
```diff
@@ -0,0 +1,192 @@
...
+import ai.chronon.spark.test.TestUtils.makeDf
+import ai.chronon.spark.{IncompatibleSchemaException, SparkSessionBuilder, TableUtils}
+import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.{AnalysisException, DataFrame, Row, SparkSession}
+import org.junit.Assert.{assertEquals, assertTrue}
+import org.junit.Test
+
+import scala.util.Try
+
+class TableUtilsFormatTest {
```
pengyu-hou, line 14: @piyush-zlai If this test specific to delta lake?

Is it reasonable to add other formats like Hive or Iceberg to the CI?

mears-stripe (approved):

pengyu-hou (approved):
Thanks for the changes. 
Looking forward to more changes from zipline-ai! 

thankyou

PR #872, opened 2024-10-24
[Metadata]Support derived feature in GroupBy Metadata Export

yuli-han:
## Summary

We want to export feature after derivation for groupBy metadata export. This PR updates the aggMetadata generation logic for groupBy like

1. Use output schema for groupBys with derivation and groupBys without aggregation.
2. Use aggregated column for groupBys without derivation and with aggregation defined.  

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @SophieYu41 @hzding621

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -180,7 +180,9 @@ class Analyzer(tableUtils: TableUtils,
                         aggPart.inputColumn.toLowerCase)
   }
 
-  private def toAggregationMetadata(columnName: String, columnType: DataType, hasDerivation: Boolean = false): AggregationMetadata = {
+  private def toAggregationMetadata(columnName: String,
+                                    columnType: DataType,
+                                    hasDerivation: Boolean = false): AggregationMetadata = {
     val operation = if (hasDerivation) "Derivation" else "No Operation"
     AggregationMetadata(columnName, columnType, operation, "Unbounded", columnName)
   }
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -172,7 +172,7 @@ class Analyzer(tableUtils: TableUtils,
 
   }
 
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -247,11 +247,8 @@ class Analyzer(tableUtils: TableUtils,
            |""".stripMargin)
     }
 
-    val aggMetadata = if (groupByConf.aggregations != null) {
```
yuli-han, line 250: @pengyu-hou @SophieYu41 Do you know why we add this line here? Currently we are not exporting the derived groupBy feature due to this condition.

hzding621: This looks like a bug to me. Your change makes more sense. 

Can you add some test cases:
- groupBy without aggregation or derivation
- groupBy with aggregation but w/o derivation
- groupBy with aggregation and derivation
- groupBy without aggregation but with derivation

hzding621 (approved):

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -247,11 +250,12 @@ class Analyzer(tableUtils: TableUtils,
...
-      groupBy.aggPartWithSchema.map { entry => toAggregationMetadata(entry._1, entry._2) }.toArray
+    val aggMetadata = if (groupByConf.hasDerivations || groupByConf.aggregations == null) {
```
pengyu-hou, line 253: @yuli-han 
should it be without the derivations for this condition ?
```suggestion
    val aggMetadata = if (!groupByConf.hasDerivations || groupByConf.aggregations == null) {
```

pengyu-hou (approved):
One minor question, otherwise LGTM

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
yuli-han: Hi @pengyu-hou it should be with derivation right? Because in derivation case we cannot get the derived feature name from aggregation so have to pull the column name from the output schema.

PR #873, opened 2024-11-07
Add a Chronon service module to add support for serving features

piyush-zlai:
## Summary

This PR adds a module to the project to spin up a service that wraps the Fetcher code and allows us to serve features. This module leverages the [Vert.x](https://vertx.io/) project. Vert.x is a high performance service toolkit that allows us to build http / grpc services - as the serving layer can be on the hot path for some use-cases, the high perf nature of Vert.x will hopefully be a good starting point. Vert.x core is Java based and given we have 3 Scala versions to support in the Chronon project, I've chosen the path of writing the service in Java rather than dealing with the interop and the multi-version compat story for now.

The service as such adds a couple of HTTP endpoints that allow us to serve groupBys and joins (each returning a json payload). We can extend things in the future with an equivalent set of grpc endpoints if needed. Endpoints look like:
```
/ping -> Health check
/config -> Returns the svc config
/v1/features/join/<joinName> -> Bulk get request for a given join
/v1/features/groupby/<groupByName> -> Bulk get request for a given group by
```

## Why / Goal

Currently users getting going on the project have to take on the additional work of writing a feature serving layer of their own (or rely on a set of disparate clients calling and using the Fetcher libs directly). This makes it painful to set up and this burden is exacerbated for users that don't natively work on JVM languages. Hooking up the fetcher code in other languages is non-trivial as it requires rewriting a lot of the core fetcher logic and maintaining parity across as things change over time. Including a service module within the project aims to lower that barrier of entry as it allows users to package the service and have their non-JVM services hit it and get feature data over the wire.
 
## Test Plan

- [X] Added Unit Tests
- [ ] Covered by existing CI
- [X] Integration tested

(Readme in the service module directory has some details / instructions on how you can spin up the project and test against the quickstart mongo api)

## Checklist
- [ ] Documentation update

## Reviewers
@caiocamatta-stripe @mickjermsurawong-openai

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -400,7 +400,7 @@ lazy val flink = (project in file("flink"))
   )
 
 lazy val service = (project in file("service"))
-  .dependsOn(online)
+  .dependsOn(online.%("compile->compile;test->test"))
   .settings(
     assembly / assemblyJarName := s"${name.value}-${version.value}.jar",
     assembly / artifact := {
--- a/build.sbt
+++ b/build.sbt
@@ -409,6 +409,7 @@ lazy val service = (project in file("service"))
     },
     addArtifact(assembly / artifact, assembly),
     publishSettings,
+    crossScalaVersions := supportedVersions,
     libraryDependencies ++= Seq(
       "io.vertx" % "vertx-core" % "4.5.10",
       "io.vertx" % "vertx-web" % "4.5.10",
--- a/build.sbt
+++ b/build.sbt
@@ -402,6 +402,13 @@ lazy val flink = (project in file("flink"))
 lazy val service = (project in file("service"))
   .dependsOn(online)
   .settings(
+    assembly / assemblyJarName := s"${name.value}-${version.value}.jar",
+    assembly / artifact := {
+      val art = (assembly / artifact).value
+      art.withClassifier(Some("assembly"))
+    },
+    addArtifact(assembly / artifact, assembly),
+    publishSettings,
     libraryDependencies ++= Seq(
       "io.vertx" % "vertx-core" % "4.5.10",
       "io.vertx" % "vertx-web" % "4.5.10",
--- a/build.sbt
+++ b/build.sbt
@@ -420,6 +420,8 @@ lazy val service = (project in file("service"))
       // use mockito 4.x as Chronon builds on Java8
       "org.mockito" % "mockito-core" % "4.11.0" % Test,
       "io.vertx" % "vertx-unit" % "4.5.10" % Test,
+      // add codegen dep to help with mockito errors
+      "io.vertx" % "vertx-codegen" % "4.5.10" % Test,
     ),
     // Assembly settings
     assembly / assemblyJarName := s"${name.value}-${version.value}.jar",
--- a/build.sbt
+++ b/build.sbt
@@ -406,8 +406,8 @@ lazy val service = (project in file("service"))
       "io.vertx" % "vertx-core" % "4.5.10",
       "io.vertx" % "vertx-web" % "4.5.10",
       "io.vertx" % "vertx-config" % "4.5.10",
-      "ch.qos.logback" % "logback-classic" % "1.5.6",
-      "org.slf4j" % "slf4j-api" % "2.0.12",
+      "ch.qos.logback" % "logback-classic" % "1.2.11",
+      "org.slf4j" % "slf4j-api" % "1.7.36",
       "com.typesafe" % "config" % "1.4.3",
       // force netty versions -> without this we conflict with the versions pulled in from
       // our online module's spark deps which causes the web-app to not serve up content
--- a/build.sbt
+++ b/build.sbt
@@ -414,8 +414,8 @@ lazy val service = (project in file("service"))
       "io.netty" % "netty-all" % "4.1.111.Final",
       "junit" % "junit" % "4.13.2" % Test,
       "com.novocode" % "junit-interface" % "0.11" % Test,
+      "org.mockito" % "mockito-core" % "5.12.0" % Test,
       "io.vertx" % "vertx-unit" % "4.5.10" % Test,
-      "org.mockito" % "mockito-core" % "3.4.0" % Test
     ),
     // Assembly settings
     assembly / assemblyJarName := s"${name.value}-${version.value}.jar",
--- a/build.sbt
+++ b/build.sbt
@@ -417,7 +417,8 @@ lazy val service = (project in file("service"))
       "io.micrometer" % "micrometer-registry-statsd" % "1.13.6",
       "junit" % "junit" % "4.13.2" % Test,
       "com.novocode" % "junit-interface" % "0.11" % Test,
-      "org.mockito" % "mockito-core" % "5.12.0" % Test,
+      // use mockito 4.x as Chronon builds on Java8
+      "org.mockito" % "mockito-core" % "4.11.0" % Test,
       "io.vertx" % "vertx-unit" % "4.5.10" % Test,
     ),
     // Assembly settings
--- a/build.sbt
+++ b/build.sbt
@@ -412,6 +412,10 @@ lazy val service = (project in file("service"))
       // force netty versions -> without this we conflict with the versions pulled in from
       // our online module's spark deps which causes the web-app to not serve up content
       "io.netty" % "netty-all" % "4.1.111.Final",
+      "junit" % "junit" % "4.13.2" % Test,
+      "com.novocode" % "junit-interface" % "0.11" % Test,
+      "io.vertx" % "vertx-unit" % "4.5.10" % Test,
+      "org.mockito" % "mockito-core" % "3.4.0" % Test
     ),
     // Assembly settings
     assembly / assemblyJarName := s"${name.value}-${version.value}.jar",
--- a/build.sbt
+++ b/build.sbt
@@ -80,7 +80,7 @@ enablePlugins(GitVersioning, GitBranchPrompt)
 lazy val supportedVersions = List(scala211, scala212, scala213)
 
 lazy val root = (project in file("."))
-  .aggregate(api, aggregator, online, spark_uber, spark_embedded)
+  .aggregate(api, aggregator, online, spark_uber, spark_embedded, service)
   .settings(
     publish / skip := true,
     crossScalaVersions := Nil,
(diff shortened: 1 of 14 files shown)


caiocamatta-stripe:
Hey @piyush-zlai - could you comment on why Vert.x vs alternatives?

piyush-zlai:
@caiocamatta-stripe - called out in the PR description but I'll expand more:
* Vertx is a fairly high performance web service framework. A few benchmarks have found it to be better than others such as Play - [Link](https://leondesilva.medium.com/spring-vs-vert-x-vs-ktor-vs-play-framework-the-performance-battle-for-a-jvm-based-api-gateway-6eebbbd4d1b5), [Link](https://www.techempower.com/benchmarks/#section=data-r22&hw=ph&test=composite).
* Play is Scala native but the challenge is that as Chronon is using a bunch of different older Scala versions - 2.10, 2.11 and 2.12, we need to deal with this licensing issue - [Licensing](https://www.playframework.com/documentation/3.0.x/General). Support for older Scala versions requires us to use the Akka fork of Play (Play 2.9.x). If Chronon were on Scala 2.13 we could go with Play 3.x (which is on Pekko). 
* Vertx being a Java based framework also makes life easier as we don't need to deal with building / cross compiling across 3 Scala versions. 
* Vertx has a fairly rich set of docs and pretty good support for [tracing](https://vertx.io/docs/vertx-opentelemetry/java/), [testing](https://vertx.io/docs/vertx-unit/java/), [grpc](https://vertx.io/docs/vertx-grpc-netty/java/) etc so the rails from a web service framework are fairly extensible.

build.sbt
```diff
@@ -391,6 +391,60 @@ lazy val flink = (project in file("flink"))
...
+      art.withClassifier(Some("assembly"))
+    },
+    addArtifact(assembly / artifact, assembly),
+    publishSettings,
+    libraryDependencies ++= Seq(
+      "io.vertx" % "vertx-core" % "4.5.10",
+      "io.vertx" % "vertx-web" % "4.5.10",
+      "io.vertx" % "vertx-config" % "4.5.10",
+      "ch.qos.logback" % "logback-classic" % "1.2.11",
+      "org.slf4j" % "slf4j-api" % "1.7.36",
```
caiocamatta-stripe, line 409: For my own learning/curiosity, any comment on why logback instead of log4j2?

service/README.md
```diff
@@ -0,0 +1,76 @@
+# Chronon Feature Service
```
caiocamatta-stripe, line 1: Should we explicitly document the endpoints available here?

service/src/main/java/ai/chronon/service/ApiProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Optional;
+
+/**
+ * Responsible for loading the relevant concrete Chronon Api implementation and providing that
+ * for use in the Web service code. We follow similar semantics as the Driver to configure this:
+ * online.jar - Jar that contains the implementation of the Api
+ * online.class - Name of the Api class
+ * online.api.props - Structure that contains fields that are loaded and passed to the Api implementation
+ * during instantiation to configure it (e.g. connection params)
+ */
```
caiocamatta-stripe, line 21: I really appreciate the comments before the classes!

service/README.md
caiocamatta-stripe, line 1: Do you think it's worth documenting the existence of this service in the chronon website?

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
```diff
@@ -0,0 +1,124 @@
...
+    private final EntityType entityType;
+    private final JavaFetcher fetcher;
+
+    public FeaturesHandler(EntityType entityType, JavaFetcher fetcher) {
+        this.entityType = entityType;
+        this.fetcher = fetcher;
+    }
+
+    @Override
+    public void handle(RoutingContext ctx) {
```
caiocamatta-stripe, line 58: Also just for my own knowledge, will vertx use one thread per request it receives? how does it handle concurrent requests?

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
```diff
@@ -0,0 +1,124 @@
...
+                entityType.equals(EntityType.GroupBy) ? fetcher.fetchGroupBys(requests) : fetcher.fetchJoin(requests);
+        // wrap the Java future we get in a Vert.x Future to not block the worker thread
+        Future<List<JTry<Map<String, Object>>>> maybeFeatureResponses =
+                Future.fromCompletionStage(resultsJavaFuture)
+                      .map(result -> result.stream().map(FeaturesHandler::responseToMap)
+                      .collect(Collectors.toList()));
+
+        maybeFeatureResponses.onSuccess(resultList -> {
+            // as this is a bulkGet request, we might have some successful and some failed responses
+            // we return the responses in the same order as they come in and mark them as successful / failed based
```
caiocamatta-stripe, line 83: Is order the only way of knowing what response maps to what request in the bulkget call? (still figuring out what the requests and responses look like)

caiocamatta-stripe (approved):
@piyush-zlai thanks for the additional commentary on why Vertx. 

Overall this looks good to me! This being a fairly large (albeit safe) change that adds new dependencies, you may want to get reviews from a least a couple other people. 

One final thing that still wasn't clear to me - how does the service get the schemas for GroupBys it's serving? Or is that purely the Fetcher's responsibility?

piyush-zlai:
@caiocamatta-stripe thanks for taking a look. I'll tag some folks in the coming days (feel free to loop in others at Stripe as well if you'd like more eyes on this).

> One final thing that still wasn't clear to me - how does the service get the schemas for GroupBys it's serving? Or is that purely the Fetcher's responsibility?

Yeah this is done as part of the GroupByUpload jobs and the MetadataUpload (join schemas) - https://chronon.ai/getting_started/Tutorial.html#uploading-data (you need to have these upload to the same KV store as you're using in your fetcher)

If you have those two sets of jobs then your KV store is primed with GroupByUploads (which contains one row which is your GroupByServingInfo data) and the join schema.

service/README.md
piyush-zlai: I thought about this but I expect the docs to get outdated soon. A better approach might be something like wiring up [OpenApi / Swagger annotations](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations). Could tackle this in a follow up.

piyush-zlai: Yeah def worth it. I was thinking of letting it bake / clearing out some nits etc and putting up a doc update PR in a follow up.

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
piyush-zlai: Yeah the order matches exactly. Similar pattern to other bulkget apis out there (e.g. [fb graph bulkget](https://developers.facebook.com/docs/graph-api/batch-requests) ).
This looks like:
Request - /v1/features/join/quickstart/training_set.v2
```
[{"user_id": "5"}, {"user_id": "6"}, {"user_id": "7"}]
```

Response:
```
{
  "results": [
      {"status": "Success", "features": { ...},
      {"status": "Success", "features": { ...},
      {"status": "Failure", "error": "Some exception",
  ]
}
```

piyush-zlai: (I'd imagine calls would most often just be for a single element. I've wired it up as a bulkGet primarily cause the fetcher api is a bulk api and instead of sticking to a list of one it was easier to have things set up this way)

service/README.md
pengyu-hou: Plus one to this.

service/src/main/java/ai/chronon/service/WebServiceVerticle.java
```diff
@@ -0,0 +1,84 @@
...
+            ctx.json("Pong!");
+        });
+
+        // Add route to show current configuration
+        router.get("/config").handler(ctx -> {
+            ctx.response()
+               .putHeader("content-type", "application/json")
+               .end(configJsonString);
+        });
+
```
mickjermsurawong-openai, line 48: Should we expose a protected method like `addToRouter` so subclass can compose more endpoints?

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
piyush-zlai: Added a section to the README showing an example of this.

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public FeaturesHandler(EntityType entityType, JavaFetcher fetcher) {
+        this.entityType = entityType;
+        this.fetcher = fetcher;
+    }
+
+    @Override
+    public void handle(RoutingContext ctx) {
+        String entityName = ctx.pathParam("name");
+        logger.info("Retrieving {} - {}", entityType.name(), entityName);
```
pengyu-hou, line 71: should this be on debug level? 
```suggestion
        logger.debug("Retrieving {} - {}", entityType.name(), entityName);
```

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Concrete implementation of the GetFeatures endpoints. Supports loading groupBys and joins.
+ * Some notes on this:
+ * We currently support bulkGet lookups against a single groupBy / join. Attempts to lookup n different GroupBys / Joins
+ * need to be split up into n different requests.
+ * A given bulkGet request might result in some successful lookups and some failed ones. We return a 4xx or 5xx response
+ * if the overall request fails (e.g. we're not able to parse the input json, Future failure due to Api returning an error)
+ * Individual failure responses will be marked as 'Failed' however the overall response status code will be successful (200)
+ * The response list maintains the same order as the incoming request list.
```
pengyu-hou, line 35: 👍

build.sbt
```diff
@@ -399,6 +399,60 @@ lazy val flink = (project in file("flink"))
                                        "flink")
   )
 
+lazy val service = (project in file("service"))
```
pengyu-hou, line 402: thanks for the separation so that our build will not be affected by the new deps.

pengyu-hou (approved):
Thanks for the great work.

service/src/main/java/ai/chronon/service/handlers/FeaturesHandler.java
piyush-zlai: yeah lets drop down, else it will be noisy

build.sbt
caiocamatta-stripe: I'm happy with logback

PR #877, opened 2024-11-18
Skipping the timestamp check for permission failures

pallavia7:
## Summary

Moved the permission check before timestamp check. If permission fails, skipped the timestamp check

## Why / Goal

User reported an issue in analyzer: the job failed open when users don't have permission to certain Hive tables. The correct behavior should be that any table permission issues should be caught in the analyzer step. 
The root cause is that the timestamp check is enabled by default, and it runs before the table permission check. Since timestamp check requires accessing data, it failed open. 

## Test Plan

[+ ] Added Unit Tests

## Checklist
- [ ] Documentation update

## Reviewers

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208513316458489

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -287,9 +287,22 @@ class Analyzer(tableUtils: TableUtils,
       (analysis, leftDf)
     }
 
+    val gbTables = ListBuffer[String]()
+    joinConf.joinParts.toScala.foreach { part =>
+      gbTables ++= part.groupBy.sources.toScala.map(_.table)
+    }
+
+    val noAccessTables = if (validateTablePermission) {
+      runTablePermissionValidation((gbTables.toList ++ List(joinConf.left.table)).toSet)
+    } else Set()
+
     if (!skipTimestampCheck) {
-      val timestampChecks = runTimestampChecks(leftDf)
-      validateTimestampChecks(timestampChecks, "Join", name)
+      if (noAccessTables.isEmpty) {
+        val timestampChecks = runTimestampChecks(leftDf)
+        validateTimestampChecks(timestampChecks, "Join", name)
+      } else {
+        logger.warn(s"skipping the timestamp checks due to permission errors in ${noAccessTables.size} tables")
+      }
     }
 
     val leftSchema = leftDf.schema.fields
@@ -297,7 +310,6 @@ class Analyzer(tableUtils: TableUtils,
       .toMap
     val joinIntermediateValuesMetadata = ListBuffer[AggregationMetadata]()
     val keysWithError: ListBuffer[(String, String)] = ListBuffer.empty[(String, String)]
-    val gbTables = ListBuffer[String]()
     val gbStartPartitions = mutable.Map[String, List[String]]()
     // Pair of (table name, group_by name, expected_start) which indicate that the table no not have data available for the required group_by
     val dataAvailabilityErrors: ListBuffer[(String, String, String)] = ListBuffer.empty[(String, String, String)]
@@ -315,7 +327,7 @@ class Analyzer(tableUtils: TableUtils,
                        part.fullPrefix,
                        includeOutputTableName = true,
                        enableHitter = enableHitter,
-                       skipTimestampCheck = skipTimestampCheck)
+                       skipTimestampCheck = skipTimestampCheck || noAccessTables.nonEmpty)
       joinIntermediateValuesMetadata ++= aggMetadata.map { aggMeta =>
         AggregationMetadata(part.fullPrefix + "_" + aggMeta.name,
                             aggMeta.columnType,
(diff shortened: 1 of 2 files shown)


pengyu-hou:
@pallavia7 you could run `sbt scalafmt` for the CI failure.

spark/src/test/scala/ai/chronon/spark/test/AnalyzerTest.scala
```diff
@@ -338,6 +340,43 @@ class AnalyzerTest {
...
+    when(tableUtils.checkTablePermission(any(), any())).thenReturn(false)
+    val namespace = "analyzer_test_ns" + "_" + Random.alphanumeric.take(6).mkString
+    tableUtils.createDatabase(namespace)
+    // left side
+    // create the event source with values
+    getTestGBSourceWithTs(namespace=namespace)
+
+    // join parts
+    val joinPart = Builders.GroupBy(
+      sources = Seq(getTestGBSourceWithTs(namespace=namespace)),
```
pengyu-hou, line 357: I am not quite sure how access denial is replayed here. Could you explain? Thanks

pallavia7: I used Mockito.spy on tableUtils objects.  when(tableUtils.checkTablePermission(any(), any())).thenReturn(false) modifies checkPermission behavior. So when tableUtils.checkTablePermission is invoked, the method returns false which happens when there is no access to table.

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -287,7 +287,16 @@ class Analyzer(tableUtils: TableUtils,
...
+    val gbTables = ListBuffer[String]()
+    joinConf.joinParts.toScala.foreach { part =>
+      gbTables ++= part.groupBy.sources.toScala.map(_.table)
+    }
+
+    val noAccessTables = if (validateTablePermission) {
+      runTablePermissionValidation((gbTables.toList ++ List(joinConf.left.table)).toSet)
+    } else Set()
+
+    if (!skipTimestampCheck && noAccessTables.isEmpty) {
```
hzding621, line 299: can we print a warning error if we are skipping the timestamp check because there are tables with permission issues?

spark/src/test/scala/ai/chronon/spark/test/AnalyzerTest.scala
```diff
@@ -338,6 +340,43 @@ class AnalyzerTest {
 
   }
 
+
+  @Test(expected = classOf[java.lang.AssertionError])
+  def testJoinAnalyzerInvalidTablePermissions(): Unit = {
+    val spark: SparkSession = SparkSessionBuilder.build("AnalyzerTest" + "_" + Random.alphanumeric.take(6).mkString, local = true)
+    val tableUtils = spy(TableUtils(spark))
+    when(tableUtils.checkTablePermission(any(), any())).thenReturn(false)
```
hzding621, line 348: consider mocking the `tableUtils.sql` to throw a runtime exception to mimic table permission, since the timestamp check logic will try to access the table data using `tableUtils.sql`, and we want to ensure that this is not triggered and gated by the permission check first.

pallavia7: I did not find a way to mock tableUtils.sql, as some sql queries like create etc we want to behave as existing and only certain sql statements to mock. So, I made sure that timestamp check logic function is not called by spying on mocked function with verify(analyzer, never()).runTimestampChecks(any(), any())

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
pallavia7: I have added warning message

hzding621 (approved):

spark/src/test/scala/ai/chronon/spark/test/AnalyzerTest.scala
```diff
@@ -338,6 +340,48 @@ class AnalyzerTest {
...
+      metaData = Builders.MetaData(name = "test_join_analyzer.key_validation", namespace = namespace, team = "chronon")
+    )
+
+    //run analyzer an ensure ts timestamp values result in analyzer passing
+    val analyzer = spy(new Analyzer(tableUtils, joinConf, oneMonthAgo, today, enableHitter = true))
+    try {
+      verify(analyzer, never()).runTimestampChecks(any(), any())
+    } catch {
+      case e: AssertionError =>
+        println("Caught unexpected AssertionError: " + e.getMessage)
```
pengyu-hou, line 380: Nit, let's be consistent with logger. 
```suggestion
        log.error("Caught unexpected AssertionError: " + e.getMessage)
```

pengyu-hou (approved):
LGTM

PR #888, opened 2024-11-27
Replace multiple calls to `withColumn` with single `select` to simplify query plans

earangol-stripe:
## Summary

Refactors some of the join code to avoid multiple calls to `withColumn` and `withColumnRenamed`.

## Why / Goal

Performance. Multiple calls to `withColumn` (and `withColumnRenamed`) [are known to cause performance issues](https://medium.com/@manuzhang/the-hidden-cost-of-spark-withcolumn-8ffea517c015) when done on too many columns, as it may generate very complex query plans. At Stripe, some of the jobs with too many columns on the RHS (e.g. ~2k) may fail due to `StackOverflowError`s on the driver when generating the query plan.

The following code snippet illustrates the issue, given a DF with 2k columns:

```scala
val range = 0 until 2000
val columns = range.map(i => s"column$i")
val schema = columns.map(c => StructField(c, IntegerType, true))
val df = spark.createDataFrame(spark.sparkContext.emptyRDD[Row], StructType(schema))
```

On a notebook, this code took ~3 minutes to run:

```scala
val renamedDf = df.columns.foldLeft(df) {
  case (partialDf, c) =>
    df.withColumnRenamed(c, s"renamed_$c")
}
renamedDf.explain() // slow
```

While this code took ~1 second:

```scala
val renamedColumns = df.columns.map(c => col(c).as(s"renamed_$c"))
val renamedDf = df.select(renamedColumns:_*)
renamedDf.explain() // fast
```

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update


nikhilsimha (commented):
very cool stuff!! lgtm will let others stamp

spark/src/main/scala/ai/chronon/spark/Extensions.scala
```diff
@@ -154,10 +156,31 @@ object Extensions {
...
     }
 
-    def prefixColumnNames(prefix: String, columns: Seq[String]): DataFrame = {
-      columns.foldLeft(df) { (renamedDf, key) =>
-        renamedDf.withColumnRenamed(key, s"${prefix}_$key")
+    def padFields(structType: sql.types.StructType): DataFrame = {
+      val existingColumns = df.columns.toSet
+      val paddedColumns = structType
+        .filterNot(field => existingColumns.contains(field.name))
+        .map(field => lit(null).cast(field.dataType).as(field.name))
```
thomaschow, line 163: will this force non-nullable fields to be nullable?

earangol-stripe: Yes, but this is not changing existing behavior. That's how the current implementation works. I added a commit that removes the current implementation of this method (I missed it initially), you can see from the diff that [the logic is the same](https://github.com/airbnb/chronon/pull/888/files#diff-b5033868ee521b9f585222ce5c40939392df053bbca36437c1d0757f67864f9dL80).

I agree this would be more correct if it respects the "nullable" definition in the schema, probably padding with a type-specific default value instead in those cases. But that's a more involved change, so I would prefer to leave it out of scope. The purpose of this PR is just to address the performance issue, not to change existing behavior.

thomaschow: ah ok thank you for including that commit!

thomaschow:
lgtm

thomaschow (approved):

earangol-stripe:
Thanks for the review!

jbrooks-stripe (approved):

nikhilsimha (approved):
LGTM!

jbrooks-stripe (approved):

earangol-stripe:
@thomaschow @nikhilsimha -- the branch got out of date and I had to merge main again so this needs another stamp. Thanks!

thomaschow (approved):
LGTM

earangol-stripe:
Oops, forgot this PR was still around. Will try to get it merged again.

PR #893, opened 2024-12-31
Bazel builds enabled along with sbt

krisnaru:
## Summary
- Enabled bazel builds for sources, tests
- Sbt works as is
- Supported scala versions: 2.12 and 2.13
- Supported spark versions 3.1, 3.2,3.5

## Why / Goal
Goal is to improve the build environment(by sandboxing) and provide multiple spark versions

## Test Plan
Unit testing green
quick start steps green https://chronon.ai/getting_started/Tutorial.html

## Build commands
- bazel build  //service
- bazel build //online
- bazel build //api/...
- Uber jar: bazel build  //spark:spark-assembly_deploy.jar
- Build for scala version and spark version
` bazel build  --config scala_2.12 --define spark_version=3.2 //spark/...`

## Checklist
- [ ] Readme updates
- [ ] Publish maven artifacts

## Reviewers
@nikhilsimha @piyush-zlai

PR #903, opened 2025-01-16
Turn off sbt delta lake tests in CI

piyush-zlai:
## Summary

Turn off delta lake tests in CI - We are seeing these tests flake out fairly often and they result in holding up the rest of our dev loops. We are exercising testing of the table utils path already (the Hive flow). We can explore enabling these on Spark 3.5 and when we're running tests on Bazel there. Isolation and Spark 3.5 are the only differences with our internal fork's setup that come to my mind and these tests have not been flaky on our end. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,36 +78,6 @@ jobs:
                   destination: spark_warehouse.tar.gz
                   when: on_fail
 
-    # run these separately as we need a isolated JVM to not have Spark session settings interfere with other runs
-    # long term goal is to refactor the current testing spark session builder and avoid adding new single test to CI
-    "Scala 13 -- Delta Lake Format Tests":
-      executor: docker_baseimg_executor
-      steps:
-        - checkout
-        - run:
-            name: Run Scala 13 tests for Delta Lake format
-            environment:
-              format_test: deltalake
-            shell: /bin/bash -leuxo pipefail
-            command: |
-              conda activate chronon_py
-              # Increase if we see OOM.
-              export SBT_OPTS="-XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=4G -Xmx4G -Xms2G"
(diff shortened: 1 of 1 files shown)


piyush-zlai:
@pengyu-hou / @caiocamatta-stripe / @mickjermsurawong-openai  - could either of you folks review?

pengyu-hou (approved):
Thanks. Once we have the bazel [PR](https://github.com/airbnb/chronon/pull/893) merged, we should have spark 3.5. Let's revisit this later.

.circleci/config.yml
```diff
@@ -177,9 +147,6 @@ workflows:
             - "Scala 13 -- Tests":
                   requires:
                       - "Pull Docker Image"
-            - "Scala 13 -- Delta Lake Format Tests":
```
pengyu-hou, line 180: On a second thought, is it possible to make it optional? So it won't be a blocker to merge PRs.

piyush-zlai: let me check if circle ci supports this - if possible I agree, it would be nice

caiocamatta-stripe (approved):

.circleci/config.yml
abbywh: @piyush-zlai any follow up here? Working on a similar PR for iceberg https://github.com/airbnb/chronon/pull/993

piyush-zlai: @abbywh - I haven't had the chance to revisit these. Based on my (limited) understanding of Spark the way we're creating sessions per test coupled with kicking off all the tests in parallel does often result in flakes as the sessions seem to conflict. The tests pass standalone in this repo (and we have them enabled for months now in our fork) but they seem to be flaking out when turned on with other tests triggered at the same time I think.

PR #905, opened 2025-01-17
fix: reuse batch response max timestamp immediately 

hzding621:
## Summary

This change will immediately use the batch response's max timestamp in the lambda aggregation to exclude streaming rows that are before the timestamp, without waiting for groupBy serving info to get refreshed. 

Only the uncached flow is updated for now, because the (cached) batchIr objects currently don't include the timestamp, and change to this is left out of scope. 

## Why / Goal

When batch response's latest timestamp is later than GroupByServingInfo's end ts (which indicates staleness of GroupByServingInfo's staleness), the TTL cache is updated asynchronously, and occasionally can fail. In the meantime, the initial request will keep using stale end ts, causing the aggregation results to be wrong as extra streaming rows that should've been excluded is included. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

wip
## Checklist
- [ ] Documentation update

## Reviewers

@pengyu-hou @yuli-han

### DIFF
--- a/online/src/main/scala/ai/chronon/online/FetcherBase.scala
+++ b/online/src/main/scala/ai/chronon/online/FetcherBase.scala
@@ -63,7 +63,7 @@ class FetcherBase(kvStore: KVStore,
                                        totalResponseValueBytes: Int,
                                        keys: Map[String, Any] // The keys are used only for caching
   ): Map[String, AnyRef] = {
-    val servingInfo = getServingInfo(oldServingInfo, batchResponses)
+    val (servingInfo, batchResponseMaxTs) = getServingInfo(oldServingInfo, batchResponses)
 
     // Batch metrics
     batchResponses match {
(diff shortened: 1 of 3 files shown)


pengyu-hou (approved):
Nice!

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -264,11 +268,12 @@ class FetcherBase(kvStore: KVStore,
     * @return the GroupByServingInfoParsed containing the latest serving information.
```
pengyu-hou, line 268: Added one comment here for the return param 
```suggestion
    * @return the GroupByServingInfoParsed containing the latest serving information.
    * @return the maximum ts from the batch IR responses. It will be later used for filtering the streaming rows after the such ts to avoid any potential duplicate records between the batch IR and streaming rows. 
```

yuli-han (approved):

PR #909, opened 2025-01-25
Make fetcher thread pool separate and pluggable 

hzding621:
## Summary

Currently Fetcher and KVStore share the same execution context:  

```
// MetadataStore.scala
implicit val executionContext: ExecutionContext = kvStore.executionContext
```

This could lead to potential deadlock if when TTLCache (e.g. called by Fetcher => getGroupByServingInfo) is trying to initiate kvStore requests to refresh cache but the thread pool is already exhausted. 

This PR 
* updates executionContext in MetadataStore (which FetcherBase/Fetcher extends) to a separate one
* allows pass in from Fetcher initialization so can be customized by users. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@pengyu-hou @pkundurthy 
cc @sup @nikhil-zlai thx both for the help investigating this

### DIFF
--- a/online/src/main/scala/ai/chronon/online/Fetcher.scala
+++ b/online/src/main/scala/ai/chronon/online/Fetcher.scala
@@ -35,7 +35,7 @@ import java.util.function.Consumer
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
 import scala.collection.{Seq, mutable}
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Random, Success, Try}
 
 object Fetcher {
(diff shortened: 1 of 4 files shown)


online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -85,8 +85,15 @@ class Fetcher(val kvStore: KVStore,
               val externalSourceRegistry: ExternalSourceRegistry = null,
               callerName: String = null,
               flagStore: FlagStore = null,
-              disableErrorThrows: Boolean = false)
-    extends FetcherBase(kvStore, metaDataSet, timeoutMillis, debug, flagStore, disableErrorThrows) {
+              disableErrorThrows: Boolean = false,
+              executionContextOverride: ExecutionContext = null)
```
sup, line 89: Thanks for making this customizable too!

Maybe the parameter should be an Option from the start to avoid nulls? Though I don't feel too strongly about that since the rest of the fields aren't Options either for symmetry.

hzding621: Yeah, I think everything is using null here mainly to simplify converting from inputs from JavaFetcher, JavaRequest, etc.

sup: Ah makes sense 👍

pengyu-hou (approved):

nikhil-zlai (approved):
Very curious,  were you able to test to verify if this helps with thread starvation issues?

PR #913, opened 2025-01-28
Set team name for inline modules

xiaohui-sun:
## Summary

Adjust team name for inline modules.
For inline modules, like a GroupBy inside a Join, we need to set the correct team name for the GroupBy using the module name.

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621

### DIFF
--- a/api/py/ai/chronon/repo/extract_objects.py
+++ b/api/py/ai/chronon/repo/extract_objects.py
@@ -18,6 +18,7 @@ import importlib.util
 import logging
 import os
 
+from ai.chronon.api.ttypes import GroupBy, Join
 from ai.chronon.logger import get_logger
 
 
@@ -40,14 +41,7 @@ def from_folder(root_path: str, full_path: str, cls: type, log_level=logging.INF
     return result
 
 
-def import_module_set_all_teams(module):
-    for name, obj in list(module.__dict__.items()):
-        if hasattr(obj, "metaData") and hasattr(obj.metaData, "team"):
-            obj.metaData.team = module.__name__.split(".")[1]
-    return module
-
-
-def import_module_set_name(module, cls):
+def import_module_set_name(module, cls, inline_group_bys=None):
     """evaluate imported modules to assign object name"""
     for name, obj in list(module.__dict__.items()):
         if isinstance(obj, cls):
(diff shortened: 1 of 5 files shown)


api/py/ai/chronon/repo/extract_objects.py
```diff
@@ -46,6 +40,13 @@ def from_folder(root_path: str,
     return result
 
 
+def import_module_set_all_teams(module):
+    for name, obj in list(module.__dict__.items()):
+        if hasattr(obj, "metaData") and hasattr(obj.metaData, "team"):
+            obj.metaData.team = module.__name__.split(".")[1]
```
pkundurthy, line 46: I'm wondering if it is safer to simply overwrite a sub-objects `team` with the parent object's `team` instead of assuming that the user will have the team correctly set in the `__name__` of the object.

pkundurthy: So basically I'm suggesting that on line 59  we add this:

```{python}
  if type(obj) == Join:
      for jp in obj.joinParts:
          if jp.groupBy.metaData.team != obj.metaData.team:
              jp.groupBy.metaData.team = obj.metaData.team
```

after the 
```
obj.metaData.name = module.__name__.partition(".")[2] + "." + name
obj.metaData.team = module.__name__.split(".")[1]
```

This will trigger only for Joins with JoinParts, and make a switch when the joinPart groupBy team is not the same that of the join team. 

@hzding621 - can you confirm if this is safe to do?

pkundurthy: we can do checks like this type(obj) == Join, with the following import
```
from ai.chronon.api.ttypes import GroupBy, Join
```

hzding621: > instead of assuming that the user will have the team correctly set in the __name__ of the object.

Actually, we can assume that `name` is always present. There is a compiler check for that: https://github.com/airbnb/chronon/blob/main/api/py/ai/chronon/join.py#L77-L78

xiaohui-sun: Thanks for the suggestions. I pushed a new iteration to use the following logic:
1. Get `inline_group_bys`, which are GroupBys when parsing Joins.
2. For these `inline_group_bys`, set the team name as the Join's team name.

pkundurthy (approved):

hzding621 (approved):

PR #919, opened 2025-02-07
fix: improve bootstrap failure error message

hzding621:
## Summary

improve error message when join job fails to parse `matched_hash` from bootstrap table, likely because bootstrap table is out dated, not cleaned up because user unset and skipped semantic_hash. 

instead of automatic recovery, for now just print a message to ask user to manually clean up the bootstrap table to be more safe.

## Why / Goal

improve error message readability

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@pallavia7

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Join.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Join.scala
@@ -155,6 +155,22 @@ class Join(joinConf: api.Join,
         }.toSeq
       }
 
+    val distinctBootstrapSetsWithSchema: Seq[(Seq[String], Long, Set[StructField])] =
+      distinctBootstrapSets.map {
+        case (hashes, rowCount) =>
+          if (hashes.exists(hash => !bootstrapInfo.hashToSchema.contains(hash))) {
+            logger.error(s"""Bootstrap table contains out-of-date metadata and should be cleaned up.
+                 |This is most likely caused by bootstrap table not properly archived during schema evolution. The semantic_hash may have been manually cleared to skip this.
+                 |Please remove old data from bootstrap table for this partition range.
+                 |
+                 |${leftRange.partitions.map(ds => s"ALTER TABLE ${bootstrapTable} DROP IF EXISTS PARTITION (ds='${ds}')").mkString("\n")}
+                 |""".stripMargin)
+            throw new IllegalStateException("Bootstrap table contains out-of-date metadata and should be cleaned up.")
+          }
+          val schema = hashes.toSet.flatMap(bootstrapInfo.hashToSchema.apply)
+          (hashes, rowCount, schema)
+      }
+
     val partsToCompute: Seq[JoinPartMetadata] = {
       if (selectedJoinParts.isEmpty) {
         bootstrapInfo.joinParts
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/Join.scala
```diff
@@ -155,6 +155,22 @@ class Join(joinConf: api.Join,
...
 
+    val distinctBootstrapSetsWithSchema: Seq[(Seq[String], Long, Set[StructField])] =
+      distinctBootstrapSets.map {
+        case (hashes, rowCount) =>
+          if (hashes.exists(hash => !bootstrapInfo.hashToSchema.contains(hash))) {
+            logger.error(s"""Bootstrap table contains out-of-date metadata and should be cleaned up.
+                 |This is most likely caused by bootstrap table not properly archived during schema evolution. The semantic_hash may have been manually cleared to skip this.
+                 |Please remove old data from bootstrap table for this partition range.
+                 |
+                 |${leftRange.partitions.map(ds => s"ALTER TABLE ${bootstrapTable} DROP IF EXISTS PARTITION (ds='${ds}')").mkString("\n")}
```
xiaohui-sun, line 166: Since hash is table level, should user just delete the table instead of dropping a partition?

hzding621: yeah I think that's also a fair solution. There is an edge scenario where dropping partitions is more efficient. If we keep older partitions, we can avoid recomputing them. It only works if older partitions of the output table is also already computed.

hzding621: honestly it's not a big deal. archive table is simpler and also easier to run

pallavia7 (approved):

pengyu-hou (approved):

spark/src/main/scala/ai/chronon/spark/Join.scala
```diff
@@ -155,6 +155,22 @@ class Join(joinConf: api.Join,
...
 
+    val distinctBootstrapSetsWithSchema: Seq[(Seq[String], Long, Set[StructField])] =
+      distinctBootstrapSets.map {
+        case (hashes, rowCount) =>
+          if (hashes.exists(hash => !bootstrapInfo.hashToSchema.contains(hash))) {
+            logger.error(s"""Bootstrap table contains out-of-date metadata and should be cleaned up.
+                 |This is most likely caused by bootstrap table not properly archived during schema evolution. The semantic_hash may have been manually cleared to skip this.
+                 |Please manually archive the old bootstrap table and re-run.
+                 |
+                 |ALTER TABLE ${bootstrapTable} RENAME TO ${bootstrapTable}_<timestamp_suffix>;
```
pengyu-hou, line 166: optional: 

Do we want to archive it for users rather than asking them to run it manually? Some users do not know where to run the drop commands.

hzding621: I think it's better to fail in this case rather than automating everything. The table metadata is already in an inconsistent state so it's better to recover things manually.

PR #930, opened 2025-02-26
Enable logging levels, driven by log4j config

krisnaru:
## Summary
This PR enables the logging, logging levels can be set from log4j conf

## Why / Goal
Users should be able to drive config for which they want

## Reviewers

@hzding621 @caiocamatta-stripe @nikhil-zlai


pengyu-hou (changes requested):
This will spam our logging.

lakshmanayenduri:
> This will spam our logging.

@pengyu-hou hardcoding log4j level inside code is anti pattern, it should be driven by log4j properties. I would suggest airbnb to fix your deployment pipelines to pass the log4j properties while submitting spark jobs to EMR clusters.

krisnaru:
> > This will spam our logging.
> 
> @pengyu-hou hardcoding log4j level inside code is anti pattern, it should be driven by log4j properties. I would suggest airbnb to fix your deployment pipelines to pass the log4j properties while submitting spark jobs to EMR clusters.

Thanks Laxman for review. let me work with pengyu and push this MR

pengyu-hou:
Thanks @krisnaru . I will spend sometime on it and test it from our end.

pengyu-hou:
@divyamanohar-stripe it is this PR.

pengyu-hou (approved):
Thanks for the PR!

PR #933, opened 2025-03-06
Extract lineage metadata from configs

xiaohui-sun:
### Context
Column level lineage can provide a detailed view of how raw data is transformed into features and subsequently used in ML models.  By capturing this lineage, we can build multiple data governance solutions, including column-level user data tagging, duplication detection, and tracing data quality issues etc.

This PR added a parser to generate column level lineage metadata.

### Usage
To use this parser, pass in a config path.
[Here](https://github.com/airbnb/chronon/compare/lineage?expand=1#diff-409734fb5415113572613797ab77b954fe5e67ea1240031b0bb556addcd81226R24) is an example.

### Output
The following metadata will be generated (defined in the [LineageMetaData](https://github.com/airbnb/chronon/pull/933/files#diff-ea01ce89b1f9449c1034d774b722e9439d066ec22211232f6eeca71fc5e99b41) class) from configs. 

image

1. Column-Level Lineage:
- Input Column: The source column.
- Output Column: The destination column.
- Operations: The transformations or operations applied between the input and output columns.

2. Module Metadata:
- Module Type: GroupBy, Join or StagingQuery

3. Table Metadata:

- Table Type: GroupBy tables, Join tables, Staging query tables or external tables (Chronon input tables)
- Table Name
- Key Columns: The primary or key columns.
- Table Columns: All columns in the table.
- Module Name

4. Feature Metadata:

- Feature Name
- Column Mapping: The mapping to the physical column if the feature is materialized as a Hive table.
- Module Name

### Implementation Details
Model Join and GroupBy into multiple steps, and for each step, generate virtual SQLs to represent the lineage logic, then these virtual SQLs are parsed using [sqlglot](https://github.com/tobymao/sqlglot) to get the lineage. 

The generated virtual SQLs could be found from the unit tests:
- GroupBy: [aggregate](https://github.com/airbnb/chronon/compare/lineage?expand=1#diff-7ad1651aec53ed9b2b72f8010af6b192beeb281f8e03c66f6fd683170eb2bf71), [derive](https://github.com/airbnb/chronon/compare/lineage?expand=1#diff-8b686044f8d3d54f2d867532641cd59d3078b9a1c2bb586afc38a2137a548940)
- Join: [join](https://github.com/airbnb/chronon/compare/lineage?expand=1#diff-8956025ca8aaf8bee7a0690b82a310070362bb27567754a6d4b4d7a64f0f5335), [derive](https://github.com/airbnb/chronon/compare/lineage?expand=1#diff-b77131712659ecdb7258648cbc2bc3f4e670f8ec9bb55e717fcbdcefff49ff00)

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @yuli-han


api/py/requirements/dev.txt
```diff
@@ -1,110 +1,87 @@
...
-# by the following command:
+# This file is autogenerated by pip-compile-multi
+# To update, run:
 #
-#    pip-compile requirements/dev.in
+#    pip-compile-multi
 #
-autoflake==2.1.1
+-r base.txt
+autoflake==2.3.1
```
hzding621, line 9: chronon CI uses python 3.7: https://github.com/airbnb/chronon/blob/main/.circleci/Dockerfile#L74

but looks like sqlglot brings in autoflake==2.3.1 which requires python >= 3.8

hzding621: maybe it's time to bump to 3.8 or even 3.10

xiaohui-sun: Current Airbnb's Airflow is 1.10 which only supports 3.7. Maybe we can keep it to 3.7?

api/py/ai/chronon/lineage/lineage_metadata.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        Initializes a new instance of LineageMetaData for storing data lineage information.
+
+        This class maintains:
+          - A set of lineage tuples: (input_column, output_column, combined_operations).
+          - A mapping of table names to Table objects.
+          - A mapping of feature identifiers to Feature objects.
+          - A mapping of tables to lists of unparsed column names.
+        """
+        self.lineages: Set[Tuple[str, str, str]] = set()
```
nikhil-zlai, line 73: make this tuple a dataclass?

api/py/ai/chronon/lineage/lineage_parser.py
```diff
@@ -0,0 +1,761 @@
...
+        """
+
+        self.base_path: Optional[str] = None
+        self.team_conf: Optional[Dict[str, Any]] = None
+        self.staging_query_tables: Dict[str, Any] = {}
+        self.metadata: LineageMetaData = LineageMetaData()
+
+    def parse_lineage(
+        self, base_path: str, entities: Optional[Set[str]] = None
+    ) -> Set[Tuple[Tuple[str, Tuple[str]], str]]:
```
nikhil-zlai, line 62: should the return type be LineageMetadata?

api/py/ai/chronon/lineage/lineage_parser.py
```diff
@@ -0,0 +1,761 @@
...
+        queue = [(column_lineage, [get_transform_operation(column_lineage.expression)])]
+        while queue:
+            parent, ops = queue.pop(0)
+            for child in parent.downstream:
+                if isinstance(child.expression, exp.Table):
+                    input_columns.append((_get_col_node_name(child), tuple(ops)))
+                child_op = get_transform_operation(child.expression)
+                # store operation if it is not a duplicate one
+                queue.append((child, ops + [child_op] if child_op != ops[-1] else ops))
+        parsed_lineages[f"{output_table}.{output_column}"] = input_columns
```
nikhil-zlai, line 760: i see, won't this capture the expressions or the lineage of transformations? just the raw connections between inputs & outputs?

api/py/test/lineage/test_parse_lineage.py
```diff
@@ -0,0 +1,77 @@
...
+            FROM (
+                SELECT e, f, g, h FROM input_table
+            )
+        """
+
+        lineage = build_lineage(output_table, sql, sources={})
+        expected_lineage = {
+            "output_table.a": [("input_table.f", ("Alias",))],
+            "output_table.b": [
+                ("input_table.f", ("Add", "Alias")),
```
nikhil-zlai, line 58: shouldn't the structure of this be recursive?

example - 
```
selects::{c : a +b, f: d + e} 
avg(c) as c_avg, avg(f) as f_avg 
derive(g = c_avg + f_avg)
```

should produce
```
g <- c_avg <- c <- a
                        c <- b
        f_avg <- f <- d
                       f <- e
```

nikhil-zlai: is the idea to flatten intermediate nodes?

nikhil-zlai (commented):
very cool PR!

api/py/requirements/dev.txt
nikhil-zlai: we used to containerize the python code that ran in airflow (used to be called `ZiplineDockerOperator` in the very old version). just an option if we wanted to modernize the chronon-ai package.

3.10 brings in pattern matching which is very useful for working with ASTs produced by sql glot.

https://peps.python.org/pep-0636/

api/py/ai/chronon/lineage/lineage_metadata.py
xiaohui-sun: Sure, we can do that.

api/py/ai/chronon/lineage/lineage_parser.py
xiaohui-sun: Sounds good.

api/py/test/lineage/test_parse_lineage.py
xiaohui-sun: If the intermediate output is not materialized - like features before derivation, or not an official Chronon output - like staging query output, then we don't want to capture the lineage.

Otherwise we want to capture the lineage even it is considered as "internal" to another entity. E.g, a Join output could be chained as an input of another GroupBy, but we want to capture the output.

nikhil-zlai: lets say we have a where clause somewhere filtering by discriminatory fields like age or gender, don't you want to show that in the lineage? 

there is a lot of reasons from a governance POV to preserve intermediate lineage. IIRC, there was a strong ask at my previous workplace ;-)

xiaohui-sun: I see. Seems we are talking about different lineages. 

Take the following for example, from column lineage perspective `c` is generated from `a` and `b`, and `f` is generated from `d` and `e`. Although `g` is used as a filter, it is not a direct lineage for `c` or `f`.

```
select a + b as c, d + e as f
from input
where g = 1
```

One of the usage for this lineage is to have automated personal data tagging. If a column is used as a filter but not in output, it is not a concern from personal data tagging perspective.

If needed we can parse all columns used from input tables. Then it would be `a`, `b`, `d`, `e` and `g` in this case.

api/py/requirements/dev.txt
xiaohui-sun: Double checked the `ZiplineDockerOperator` has been deprecated, and it is also for BigAir, which has the capability of specifying a customized docker image. For the airflow version used at Airbnb it doesn't support newer versions than 3.7 so we will just keep it here.

api/py/test/lineage/group_by_sqls/aggregate.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+  subject,
+  AGG(cnt) AS cnt_count,
+  AGG(event_id) AS event_id_approx_percentile,
+  AGG(event_id) AS event_id_last,
+  AGG(event_id) AS event_id_sum
```
hzding621, line 6: for now we are using AGG for all aggregation functions which is fine. 

For our CLL effort, I do see we may want to support different `OperationType` with semantics (e.g. COPY, CONCAT, etc.) I suppose this is important because `OperationType` retain raw data but others obscure it. Is there a way to extend to support such info?

api/py/test/lineage/test_parse_lineage.py
hzding621: >Take the following for example, from column lineage perspective `c` is generated from `a` and `b`, and `f` is generated from `d` and `e`. Although `g` is used as a filter, it is not a direct lineage for `c` or `f`.

```
select a + b as c, d + e as f
from input
where g = 1
```

whether `where` should be part of lineage is a great question. 

I suppose in this scenario, if we want to consider `where`s, then we would have: 

- c depends on a, b, **g**
- f depends on d, e, **f**

Support `where` can be out of scope for this PR, and extend later if need arises

hzding621 (approved):
💯 this is huge!!! great job on the lib design and tests coverage

api/py/ai/chronon/lineage/lineage_parser.py
```diff
@@ -0,0 +1,761 @@
...
+            Tuple[str, str, Dict[str, str]]:
+                - table (str): The table name.
+                - filter_expr (str): The combined filter expression.
+                - table_type (str): Type of source table.
+                - selects (Dict[str, str]): Mapping of select expressions.
+        """
+        table_type = "HIVE_TABLE"
+        if source.entities:
+            if source.entities.query and source.entities.query.mutationTimeColumn:
+                table_type = "DB_EXPORT"
```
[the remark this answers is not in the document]
xiaohui-sun: I will remove the `table_type`. Right now we don't need to differentiate a topic source from a hive table source.

api/py/ai/chronon/lineage/lineage_parser.py
```diff
@@ -0,0 +1,761 @@
...
+
+        """
+
+        self.base_path: Optional[str] = None
+        self.team_conf: Optional[Dict[str, Any]] = None
+        self.staging_query_tables: Dict[str, Any] = {}
+        self.metadata: LineageMetaData = LineageMetaData()
+
+    def parse_lineage(
+        self, base_path: str, entities: Optional[Set[str]] = None
```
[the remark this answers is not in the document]
xiaohui-sun: will do.

api/py/test/lineage/group_by_sqls/aggregate.sql
xiaohui-sun: Instead of using AGG, I can track the actual feature aggregation types here.
For other user defined/free form sql, we can keep the parsed operation sequences.

api/py/test/lineage/test_parse_lineage.py
nikhil-zlai: fwiw, openLineage models these as indirect dependencies - and seems to be able to connect lineage through the join.

xiaohui-sun: Yes it is modeled as `Indirect` [transformation type](https://openlineage.io/docs/spec/facets/dataset-facets/column_lineage_facet/) in OpenLineage. Right how we only model `Direct` transformation and it could be a future work.

nikhil-zlai: good with me.

nikhil-zlai (approved):
great work!

api/py/ai/chronon/lineage/lineage_metadata.py
```diff
@@ -0,0 +1,219 @@
...
+class Table:
+    def __init__(self, table_name: str, table_type: TableType) -> None:
+        """
+        Instance initializer for the Table class.
+
+        :param table_name: The name of the table.
+        :param table_type: The type of the table as defined in TableType.
+        """
+        self.columns: Set[str] = set()  # Set to hold column names for this table.
+        self.key_columns: Set[str] = set()  # Set to hold key column names.
```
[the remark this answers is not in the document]
xiaohui-sun: key_columns are used for joining and are candidates for personal data columns.
This is important for table level metadata - although we can add more if it is needed, like if this table is a Hive table or a topic, partition column, etc.

api/py/ai/chronon/lineage/lineage_metadata.py
```diff
@@ -0,0 +1,219 @@
...
+        self.column: Optional[str] = column
+
+
+class LineageMetaData:
+    def __init__(self) -> None:
+        """
+        Initializes a new instance of LineageMetaData for storing data lineage information.
+
+        This class maintains:
+          - A set of lineage tuples: (input_column, output_column, combined_operations).
```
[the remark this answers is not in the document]
xiaohui-sun: sounds good

api/py/ai/chronon/lineage/lineage_parser.py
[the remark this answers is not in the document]
xiaohui-sun: changed to "module_names"

api/py/test/lineage/group_by_sqls/aggregate.sql
xiaohui-sun: updated

api/py/ai/chronon/lineage/lineage_parser.py
xiaohui-sun: updated

api/py/ai/chronon/lineage/lineage_parser.py
```diff
@@ -0,0 +1,761 @@
...
+        """
+        Generate all aggregate expression names based on the aggregation's windows and buckets.
+
+        Args:
+            aggregation (Any): The aggregation configuration object.
+
+        Returns:
+            List[str]: A list of aggregate expression names.
+        """
+        base_name = f"{_get_op_suffix(aggregation.operation, aggregation.argMap)}"
```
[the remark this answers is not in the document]
xiaohui-sun: Did you mean using group_by.get_output_col_names?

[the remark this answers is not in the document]
hzding621: yep, but I think `group_by.get_output_col_names` is different in that it includes a prefix that's not included here. maybe we can parameterize it so that we can use at both places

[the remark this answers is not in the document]
xiaohui-sun: I see. We can directly use it. Removed the duplicate code.

pengyu-hou (approved):
Looks good. Thanks for adding detailed method description and comprehensive testing.

api/py/ai/chronon/lineage/lineage_parser.py
```diff
@@ -0,0 +1,756 @@
...
+                features.extend(derived_features)
+            else:
+                features = derived_features
+
+            derive_sql = self.build_gb_derive_sql("aggregate_table", gb)
+            sources["aggregate_table"] = sql
+            sql = derive_sql.sql(pretty=True)
+
+        # If join_part_table is defined, then it generates a join part table.
+        # If gb.backfill_start_date is defined, then it generates a backfill table.
```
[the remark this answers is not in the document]
xiaohui-sun: updated.

PR #935, opened 2025-03-06
Custom partition coumn for iceberg tables

krisnaru:
## Summary
There is a hardcoded dependency on the 'ds' partition column in Iceberg table integrations, preventing the use of custom partition columns specified via the Spark configuration property spark.chronon.partition.column.

## Why / Goal
The current implementation assumes 'ds' as the default partition column for Iceberg tables, which restricts flexibility in environments where different partitioning schemes are required. This limitation was highlighted in a recent discussion on the Chronon Slack channel.
Modify the Iceberg integration to dynamically determine the partition column based on the spark.chronon.partition.column configuration property. If this property is not set, default to 'ds' to maintain backward compatibility.

## Reviewers
@mickjermsurawong-openai @caiocamatta-stripe @nikhil-zlai

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
@@ -208,7 +208,7 @@ case object Iceberg extends Format {
       throw new NotImplementedError(s"subPartitionsFilter is not supported on this format")
     }
 
-    getIcebergPartitions(tableName)
+    getIcebergPartitions(tableName, partitionColumn)
   }
 
   override def partitions(tableName: String)(implicit sparkSession: SparkSession): Seq[Map[String, String]] = {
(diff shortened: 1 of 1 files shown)


nikhil-zlai (approved):
Thanks for the fix! 

Might be worth starting some iceberg local tests if you rely on iceberg in prod.

hzding621 (approved):

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -217,21 +217,22 @@ case object Iceberg extends Format {
         "For single partition retrieval, please use 'partition' method.")
   }
 
-  private def getIcebergPartitions(tableName: String)(implicit sparkSession: SparkSession): Seq[String] = {
+  private def getIcebergPartitions(tableName: String, partitionColumn: String)(implicit
+      sparkSession: SparkSession): Seq[String] = {
```
caiocamatta-stripe, line 221: Thoughts on setting `ds` as default?

caiocamatta-stripe (approved):

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
caiocamatta-stripe: actually, that's probably not a good default

PR #940, opened 2025-03-11
Fixing assert on backfillStartDate

krisnaru:
## Summary
Enable overrideStartPartition when backfillStartDate is not passed

## Why / Goal
This will help airflow DAG to run everyday without specifying it from Feature definition

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/GroupBy.scala
+++ b/spark/src/main/scala/ai/chronon/spark/GroupBy.scala
@@ -677,8 +677,10 @@ object GroupBy {
                       overrideStartPartition: Option[String] = None,
                       skipFirstHole: Boolean = true): Unit = {
     assert(
-      groupByConf.backfillStartDate != null,
-      s"GroupBy:${groupByConf.metaData.name} has null backfillStartDate. This needs to be set for offline backfilling.")
+      groupByConf.backfillStartDate != null || overrideStartPartition != null,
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/GroupBy.scala
```diff
@@ -686,8 +686,10 @@ object GroupBy {
...
                       skipFirstHole: Boolean = true): Unit = {
     assert(
-      groupByConf.backfillStartDate != null,
-      s"GroupBy:${groupByConf.metaData.name} has null backfillStartDate. This needs to be set for offline backfilling.")
+      groupByConf.backfillStartDate != null || overrideStartPartition != null,
+      s"GroupBy:${groupByConf.metaData.name} has null backfillStartDate or overrideStartPartition." +
+        s" This needs to be set for offline backfilling."
+    )
     groupByConf.setups.foreach(tableUtils.sql)
     val overrideStart = overrideStartPartition.getOrElse(groupByConf.backfillStartDate)
```
pengyu-hou, line 694: So the group by backfill start date is meant for DAG and task scheduling purpose. If there is no such parameter, we should not even enter the group by backfill mode. 

And the start partition override is implemented at line 694 already.

pengyu-hou (changes requested):

PR #941, opened 2025-03-12
Flink trigger as param to tiledjob

srenumakala:
## Summary
Currently the trigger used is `AlwaysFireOnElementTrigger` in runTiledGroupByJob. This is hardcoded and restricts use of using other custom triggers like BufferedProcessingTimeTrigger by passing to the job. Changes include making trigger as optional parameter, defaulting to AlwaysFireOnElement if not passed

## Why / Goal
Users can pass custom trigger as required to the Flink tiled jobs

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@piyush-zlai, @caiocamatta-stripe, @pengyu-hou

### DIFF
--- a/flink/src/main/scala/ai/chronon/flink/FlinkJob.scala
+++ b/flink/src/main/scala/ai/chronon/flink/FlinkJob.scala
@@ -18,6 +18,7 @@ import org.apache.flink.api.scala._
 import org.apache.flink.streaming.api.functions.async.RichAsyncFunction
 import org.apache.flink.streaming.api.windowing.assigners.{TumblingEventTimeWindows, WindowAssigner}
 import org.apache.flink.streaming.api.windowing.time.Time
+import org.apache.flink.streaming.api.windowing.triggers.Trigger
 import org.apache.flink.streaming.api.windowing.windows.TimeWindow
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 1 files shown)


piyush-zlai (approved):

flink/src/main/scala/ai/chronon/flink/FlinkJob.scala
```diff
@@ -112,8 +113,13 @@ class FlinkJob[T](eventSrc: FlinkSource[T],
...
     *
     *  The window causes a split in the Flink DAG, so there are two nodes, (1+2) and (3+4+5).
+    *
+    * @param trigger - trigger used for window emission. Defaulted to AlwaysFireOnElementTrigger which fires on every element
+    *                An alternative is BufferedProcessingTimeTrigger which buffer writes so they happen at most every X milliseconds per GroupBy & key
     */
-  def runTiledGroupByJob(env: StreamExecutionEnvironment): DataStream[WriteResponse] = {
+  def runTiledGroupByJob(
+      env: StreamExecutionEnvironment,
+      trigger: Trigger[Map[String, Any], TimeWindow] = new AlwaysFireOnElementTrigger()): DataStream[WriteResponse] = {
```
caiocamatta-stripe, line 122: Thoughts on making this a config somewhere? Over time, I imagine there will be more and more ways to tune the Flink job, so I wonder if adding them all to the function signature is the best way to go about it

caiocamatta-stripe (approved):

flink/src/main/scala/ai/chronon/flink/FlinkJob.scala
srenumakala: good idea. could be my personal preference, if the method params are atleast more than 2, then it can be moved to config objects. Will keep this in mind, next time to tune will make a config object

pengyu-hou: makes sense to make a config object if the list is getting longer.

pengyu-hou (approved):

PR #944, opened 2025-03-14
Update spark app name for all join backfill stages

hzding621:
## Summary

1. If `selectedJoinParts` mode (where we only run a subset of join_parts), add join_parts as suffix to spark application name for easier monitoring. 
2. Update spark application name for left and final stage to match the format for easier prefix-matching

## Why / Goal

For easier monitoring and analysis. 

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@pallavia7

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Driver.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Driver.scala
@@ -344,8 +344,8 @@ object Driver {
       lazy val joinConf: api.Join = parseConf[api.Join](confPath())
       override def subcommandName() =
         if (selectedJoinParts.isDefined) {
-          val parts = selectedJoinParts().mkString("_")
-          s"join_${joinConf.metaData.name}_${parts}"
+          val parts = selectedJoinParts().mkString(",")
+          s"join_${joinConf.metaData.name}_jp_${parts}"
         } else {
           s"join_${joinConf.metaData.name}"
         }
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/Driver.scala
```diff
@@ -342,7 +342,13 @@ object Driver {
           default = Some(false),
           descr = "Whether or not to use the cached bootstrap table as the source - used in parallelized join flow.")
       lazy val joinConf: api.Join = parseConf[api.Join](confPath())
-      override def subcommandName() = s"join_${joinConf.metaData.name}"
+      override def subcommandName() =
+        if (selectedJoinParts.isDefined) {
+          val parts = selectedJoinParts().mkString(",")
```
pkundurthy, line 347: would `,` in an application name be an issue?

hzding621: @pkundurthy `,` is allowed (tested). It's pretty common actually. For example, lots of our jobs have this convention:

```
DS:2025-03-10, ID: com.company.class.XYZJob
```

pallavia7 (approved):

pengyu-hou (approved):

PR #955, opened 2025-03-28
fix: propagate kvstore multi get exception properly  

hzding621:
## Summary

Fix `KVStore.getString` such that the cases where `KVStore.multiGet` returns a `Failure` can be properly handled. 

## Why / Goal

Fetcher should handle join parts failures and return partial results. When KVStore throws exceptions for a certain GroupBy, the exception should be caught and swallowed as an exception string the responseMap instead of failing the whole request. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

@airbnb/airbnb-chronon-maintainers @pengyu-hou @yuli-han

### DIFF
--- a/online/src/main/scala/ai/chronon/online/Api.scala
+++ b/online/src/main/scala/ai/chronon/online/Api.scala
@@ -58,27 +58,29 @@ trait KVStore {
 
   // helper method to blocking read a string - used for fetching metadata & not in hotpath.
   def getString(key: String, dataset: String, timeoutMillis: Long): Try[String] = {
-    val response = getResponse(key, dataset, timeoutMillis)
-    if (response.values.isFailure) {
-      Failure(new RuntimeException(s"Request for key ${key} in dataset ${dataset} failed", response.values.failed.get))
-    } else {
-      Success(new String(response.latest.get.bytes, Constants.UTF8))
-    }
+    val bytesTry = getResponse(key, dataset, timeoutMillis)
+    bytesTry.map(bytes => new String(bytes, Constants.UTF8))
   }
 
   def getStringArray(key: String, dataset: String, timeoutMillis: Long): Try[Seq[String]] = {
-    val response = getResponse(key, dataset, timeoutMillis)
-    if (response.values.isFailure) {
-      Failure(new RuntimeException(s"Request for key ${key} in dataset ${dataset} failed", response.values.failed.get))
-    } else {
-      Success(StringArrayConverter.bytesToStrings(response.latest.get.bytes))
-    }
+    val bytesTry = getResponse(key, dataset, timeoutMillis)
+    bytesTry.map(bytes => StringArrayConverter.bytesToStrings(bytes))
   }
 
-  private def getResponse(key: String, dataset: String, timeoutMillis: Long): GetResponse = {
+  private def getResponse(key: String, dataset: String, timeoutMillis: Long): Try[Array[Byte]] = {
     val fetchRequest = KVStore.GetRequest(key.getBytes(Constants.UTF8), dataset)
     val responseFutureOpt = get(fetchRequest)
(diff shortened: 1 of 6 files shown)


yuli-han:
> When KVStore throws exceptions for a certain GroupBy, the exception should be caught and swallowed as an exception string the responseMap instead of failing the whole request.
I remembered this is already the current behavior right? Did we see any exception case?

online/src/main/scala/ai/chronon/online/Api.scala
```diff
@@ -58,27 +58,29 @@ trait KVStore {
...
-      Failure(new RuntimeException(s"Request for key ${key} in dataset ${dataset} failed", response.values.failed.get))
-    } else {
-      Success(StringArrayConverter.bytesToStrings(response.latest.get.bytes))
-    }
+    val bytesTry = getResponse(key, dataset, timeoutMillis)
+    bytesTry.map(bytes => StringArrayConverter.bytesToStrings(bytes))
   }
 
-  private def getResponse(key: String, dataset: String, timeoutMillis: Long): GetResponse = {
+  private def getResponse(key: String, dataset: String, timeoutMillis: Long): Try[Array[Byte]] = {
```
yuli-han, line 70: My understanding is we used to handle the exception in getString and getStringArray and now we handled the exception in getResponse, wander what is the missing situation for the previous solution? Also could you validate the getStringArray is working fine after the change by integration testing? The getStringArray is mainly used for fetching active join/gb list to validate whether the fetched entity has been removed.

yuli-han (approved):

pallavia7 (approved):

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -670,7 +670,7 @@ class FetcherBase(kvStore: KVStore,
                         if (debug || Math.random() < 0.001) {
                           logger.error(s"Failed to fetch $groupByRequest", ex)
                         }
-                        Map(groupByRequest.name + "_exception" -> ex.traceString)
+                        Map(prefix + "_exception" -> ex.traceString)
```
pengyu-hou, line 673: Does `prefix` have the group by's request name?

online/src/main/scala/ai/chronon/online/Api.scala
```diff
@@ -58,27 +58,29 @@ trait KVStore {
...
     val fetchRequest = KVStore.GetRequest(key.getBytes(Constants.UTF8), dataset)
     val responseFutureOpt = get(fetchRequest)
-    Await.result(responseFutureOpt, Duration(timeoutMillis, MILLISECONDS))
+    def buildException(e: Throwable) = new RuntimeException(s"Request for key ${key} in dataset ${dataset} failed", e)
+    Try(Await.result(responseFutureOpt, Duration(timeoutMillis, MILLISECONDS))) match {
+      case Failure(e) =>
+        Failure(buildException(e))
+      case Success(resp) =>
+        if (resp.values.isFailure) {
+          Failure(buildException(resp.values.failed.get))
```
pengyu-hou, line 79: oh so the failure was nested within the response value😑

Good catch.

pengyu-hou (approved):
Thanks Haozhen for fixing this corner case!

online/src/main/scala/ai/chronon/online/FetcherBase.scala
hzding621: yes, prefix here actually stores `joinPart.fullPrefix`. It's better to use `joinPart.fullPrefix` than `groupByRequest.name` because `fullPrefix` also contains `joinPart.prefix`.

PR #922, opened 2025-02-11
Fix NPE when fetching a non-exist key from a GroupBy w/ derivation

xiaohui-sun:
## Summary

There is a NPE when fetching a non-exist key from a GroupBy w/ derivation like below:
```
2025-02-11 18:07:04 ERROR Fetcher:192 - Exception Message: scala.MatchError: (Map(id_host -> 12345),null) (of class scala.Tuple2)
	at ai.chronon.online.OnlineDerivationUtil$.$anonfun$buildRenameOnlyDerivationFunction$1(OnlineDerivationUtil.scala:32)
	at ai.chronon.online.FetcherBase.$anonfun$fetchGroupBys$25(FetcherBase.scala:514)
	at scala.util.Try$.apply(Try.scala:210)
	at ai.chronon.online.FetcherBase.$anonfun$fetchGroupBys$18(FetcherBase.scala:516)
	at scala.util.Success.map(Try.scala:262)
	at ai.chronon.online.FetcherBase.$anonfun$fetchGroupBys$17(FetcherBase.scala:449)
	at scala.collection.Iterator$$anon$9.next(Iterator.scala:575)
	at scala.collection.immutable.List.prependedAll(List.scala:153)
	at scala.collection.IterableOnceOps.toList(IterableOnce.scala:1251)
	at scala.collection.IterableOnceOps.toList$(IterableOnce.scala:1251)
	at scala.collection.AbstractIterator.toList(Iterator.scala:1288)
	at ai.chronon.online.FetcherBase.$anonfun$fetchGroupBys$9(FetcherBase.scala:447)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
```

This PR fixes it by returning null values for derivations if key is not found.

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @yuli-han

PR #956, opened 2025-04-02
Revert change on FetchGroupBys return value when key doesn't exist

xiaohui-sun:
## Summary

The original change in [this PR](https://github.com/airbnb/chronon/pull/922/files#diff-9ea959844a5006c401298a8b18ac1b693a917a310345d563a10b76558127a297L490) changed the return value of fetchGroupBys when key doesn't exist.
Originally it returns Null, and after it returns an empty map.

This is an in-compatible change so we need to roll it back.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

### DIFF
--- a/online/src/main/scala/ai/chronon/online/FetcherBase.scala
+++ b/online/src/main/scala/ai/chronon/online/FetcherBase.scala
@@ -484,7 +484,7 @@ class FetcherBase(kvStore: KVStore,
               val streamingResponsesOpt =
                 streamingRequestOpt.map(responsesMap.getOrElse(_, Success(Seq.empty)).getOrElse(Seq.empty))
               val queryTs = request.atMillis.getOrElse(System.currentTimeMillis())
-              val groupByResponse: Map[String, AnyRef] =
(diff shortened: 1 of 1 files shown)


online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -539,7 +539,7 @@ class FetcherBase(kvStore: KVStore,
                 }
                 derivedMap
               } else {
-                groupByResponse
+                groupByResponse.get
```
hzding621, line 542: groupByResponse.orNull

hzding621: can we also leave a comment that returning `null` is the expected behavior for users using fetchGroupBys?

xiaohui-sun: sure

hzding621 (approved):

pengyu-hou (approved):

PR #960, opened 2025-04-05
Exclude backfill start date for versioning in join jobs

pengyu-hou:
## Summary

This PR will add a flag to exclude backfill start date in the join part in versioning check.

## Why / Goal

Exclude the backfill start date in the join part, it is intended to trigger separate group by backfill jobs, and semantic hash is only used in join for versioning check, therefore, exclude it for more flexibility of group by backfill jobs

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

### DIFF
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -907,20 +907,28 @@ object Extensions {
       join
     }
 
-    /*
-     * Compute variants of semantic_hash with different flags. A flag is stored on Hive metadata and used to
-     * indicate which version of semantic_hash logic to use.
-     */
-    def semanticHash(excludeTopic: Boolean): Map[String, String] = {
+    /**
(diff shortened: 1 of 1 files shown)


hzding621 (changes requested):
For this change we must introduce a new entry in `semantic_hash_options` because this is not backward compatible and will cause all joins that use GroupBys with backfill_start_date to have a new semantic hash and trigger an unexpected archive. You also need to handle the semantic hash migration.

pengyu-hou:
> For this change we must introduce a new entry in `semantic_hash_options` because this is not backward compatible and will cause all joins that use GroupBys with backfill_start_date to have a new semantic hash and trigger an unexpected archive. You also need to handle the semantic hash migration, similar to #801

Oh yes, thanks for the pointer. I was about to ask the PR reference. Will address it

PR #962, opened 2025-04-07
Removing system.exit as it is preventing graceful shutdown of spark s…

pallavia7:
…ometimes when run in cluster mode

## Summary
When spark is run in cluster mode, this exit call is preventing spark to gracefully exit sometimes.Removed the system.exit call. 

## Test Plan
Tested daily, backfill and analyzer jobs. Also tested failure cases when jobs fail ooms/exceptions and verified that job still finished
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [+] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209864899161513

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Driver.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Driver.scala
@@ -1062,7 +1062,6 @@ object Driver {
 
   def main(baseArgs: Array[String]): Unit = {
     val args = new Args(baseArgs)
-    var shouldExit = true
     args.subcommand match {
       case Some(x) =>
         x match {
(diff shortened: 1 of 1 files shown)


pengyu-hou (commented):
Hey, could you test it again with a streaming job? If that also works, then we can merge it.

pengyu-hou (approved):
It should be working fine for streaming mode as well. Just test it with some peace of mind.

hzding621 (approved):

pallavia7:
> It should be working fine for streaming mode as well. Just test it with some peace of mind.

I have tested streaming mode as well.

PR #963, opened 2025-04-08
Remove the special handling for iceberg tables

yuli-han:
## Summary

We used to introduce a special handling for iceberg table with hr partition. This has blocked the users from using iceberg tables with hr partition as input because chronon is not able to detect corresponding partitions from the input table. 

The PR removes the special handling and enable the multiple partitions. 

## Test Plan

Tested by integration test with iceberg hourly table. 

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@hzding621 @pengyu-hou @pallavia7 @xiaohui-sun

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
+++ b/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
@@ -172,6 +172,7 @@ abstract class JoinBase(joinConf: api.Join,
           rightRange,
           Some(Seq(joinConf.left.table)),
           inputToOutputShift = shiftDays,
+          inputTableToSubPartitionFiltersMap = Map(joinConf.left.table -> joinConf.left.subPartitionFilters),
           inputTableToPartitionColumnsMap = joinConf.left.tableToPartitionColumn,
           // never skip hole during partTable's range determination logic because we don't want partTable
           // and joinTable to be out of sync. skipping behavior is already handled in the outer loop.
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -202,49 +203,18 @@ case object Hive extends Format {
...
+      .sql(s"SHOW PARTITIONS $tableName")
+      .collect()
+      .map(row => parseHivePartition(row.getString(0)))
   }
 
   def createTableTypeString: String = "USING iceberg"
   def fileFormatString(format: String): String = ""
 
-  override def supportSubPartitionsFilter: Boolean = false
+  override def supportSubPartitionsFilter: Boolean = true
```
hzding621, line 216: Does this require changing existing configs already using hourly tables? Do we need to do any migration?

hzding621 (approved):

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
hzding621: New updated logic makes sense!

nikhil-zlai (approved):
very clean!

pengyu-hou (approved):

spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -172,6 +172,7 @@ abstract class JoinBase(joinConf: api.Join,
           rightRange,
           Some(Seq(joinConf.left.table)),
           inputToOutputShift = shiftDays,
+          inputTableToSubPartitionFiltersMap = Map(joinConf.left.table -> joinConf.left.subPartitionFilters),
```
hzding621, line 175: Actually this doesn't look correct, this is looking at join_part intermediate table, which shouldn't use `left.subPartitionFilters`

hzding621 (approved):

PR #774, opened 2024-06-14
Don't request null keys from KVStore

jbrooks-stripe:
## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

PR #968, opened 2025-04-15
Don't request from KVStore if ANY of key is null

xiaohui-sun:
## Summary

This is a copy of the original PR here: https://github.com/airbnb/chronon/pull/774 with the change:
1. Orig PR: Don't request if ALL keys are null.
2. This PR: Don't request if ANY keys are null.

The reason we want to make this change is to make offline-online consistency.
In the [offline join](https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/JoinUtils.scala#L147), if ANY keys are null then it will be empty in join.

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/py/BUILD.bazel
+++ b/api/py/BUILD.bazel
@@ -40,18 +40,21 @@ pytest_suite(
         ["test/**/*.py"],
         exclude = ["test/sample/**/*"],
     ),
-    data = glob(["test/sample/**/*"]),
+    data = glob(["test/sample/**/*",
+                 "test/lineage/**/*.sql"]),
     env = {
         "CHRONON_ROOT": "api/py/",
     },
     imports = [
         ".",
         "test/sample",
+        "chronon/api/thrift",
     ],
     deps = [
         "//api/py:api_py",
         "//api/thrift:api-models-py",
         requirement("thrift"),
         requirement("click"),
+        requirement("sqlglot")
     ],
 )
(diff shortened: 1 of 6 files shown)


online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -350,73 +350,80 @@ class FetcherBase(kvStore: KVStore,
...
-                kvStore.createKeyBytes(request.keys, groupByServingInfo, groupByServingInfo.groupByOps.batchDataset)
-              streamingKeyBytes =
-                kvStore.createKeyBytes(request.keys, groupByServingInfo, groupByServingInfo.groupByOps.streamingDataset)
-            } else throw InvalidEntityException(request.name)
-          } catch {
-            // If the group_by is inactive, throw the exception
-            case ex: InvalidEntityException => {
-              context.increment("fetch_invalid_group_by_failure.count")
+    val groupByRequestToKvRequest: Seq[(Request, Try[GroupByRequestMeta])] = requests.iterator
+      .filter(r => r.keys == null || r.keys.values == null || r.keys.values.exists(_ != null))
```
hzding621, line 354: nit: introduce another val
```
    val validRequests =
      requests.filter(r => r.keys == null || r.keys.values == null || r.keys.values.exists(_ != null))
    val groupByRequestToKvRequest: Seq[(Request, Try[GroupByRequestMeta])] = validRequests.iterator.map { request =>
```
this will avoid making deeper indentation, and also preserve git history better

xiaohui-sun: Thanks. Let me try this.

hzding621 (approved):

pengyu-hou (approved):

PR #969, opened 2025-04-15
Fix type hints for optional parameters in Python API

earangol-stripe:
## Summary

Set `Optional` type hints for parameters with default value of `None` in Python API.

## Why / Goal

To improve usability with IDEs and lint tools that (correctly) complain about setting `None` to non-optional parameters.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update


api/py/ai/chronon/query.py
```diff
@@ -82,7 +82,7 @@ def Query(
         start_partition,
         end_partition,
         time_column,
-        setups,
+        setups or [],
```
earangol-stripe, line 85: This is to prevent unexpected behaviors due to [mutable-argument-default](https://docs.astral.sh/ruff/rules/mutable-argument-default/).

jbrooks-stripe (approved):

PR #971, opened 2025-04-15
Fix type hints for enums in ttypes

earangol-stripe:
## Summary

Changes type hints for enum values to `int`s instead of the class type.

## Why / Goal

Current type hints for enum values are incorrect because of the way Thrift compiles them in Python. E.g. for `ttypes.Accuracy`:

```python
class Accuracy(object):
    TEMPORAL = 0
    SNAPSHOT = 1
```

So a parameter defined as `def my_func(accuracy: ttypes.Accuracy = ttyps.Accuracy.TEMPORAL)` ends up being an incorrect assignment because `TEMPORAL` is just an int.

To improve usability with IDEs and lint tools, along the same lines as #969.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/py/ai/chronon/group_by.py
+++ b/api/py/ai/chronon/group_by.py
@@ -33,12 +33,12 @@ LOGGER = logging.getLogger()
 
 
 def collector(
-    op: ttypes.Operation,
-) -> Callable[[ttypes.Operation], Tuple[ttypes.Operation, Dict[str, str]]]:
+    op: int,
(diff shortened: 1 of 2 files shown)


earangol-stripe:
I just realized there is an option in the Thrift compiler to generate Python enums: `py:enum`. That looks better to me, though it might not be a safe change. I'll put this PR on hold and play with that option instead.

PR #973, opened 2025-04-17
CHIP for user-provided documentation

earangol-stripe:
## Summary

Adds CHIP to support user-provided documentation.

## Why / Goal

See the CHIP.

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update


proposals/CHIP-3.md
```diff
@@ -0,0 +1,69 @@
+# CHIP-3: User-provided documentation for feature definitions
+
+https://github.com/airbnb/chronon/issues/975
+
+## Motivation
+
+Programmatic access to feature documentation is useful for integrating with systems aimed at ML explainability (e.g. [SHAP](https://shap.readthedocs.io/en/latest/)) and feature discovery (e.g. feature catalogs).
+
+Currently, it's possible to inspect Chronon definitions to determine _how_ a feature was computed, which is a type of documentation. However, there are other aspects of ML feature development that are not
+captured and documented within the feature definition itself, for example: context, domain-specific knowledge, assumptions, caveats. This CHIP aims to fill those gaps with user-provided documentation.
```
pengyu-hou, line 10: It is a valid motivation!

proposals/CHIP-3.md
```diff
@@ -0,0 +1,69 @@
...
+    xx: optional string description
+}
+```
+
+2. Add a `metaData` field to `Aggregation` and `Derivation` in the Thrift API.
+
+```thrift
+struct Aggregation {
+    ...
+    xx: optional MetaData metaData
```
pengyu-hou, line 30: The `metaData` has more fields that are irrelevant to the description itself. 

And the original thought was to keep params non related to feature calculation in the metadata, therefore, when it coms to versioning check etc, we don't have to maintain a potential increasing list of fields

earangol-stripe: > The metaData has more fields that are irrelevant to the description itself.

Yeah, this is something I noticed too, I added some comments about it in the section about alternatives. I think this could be solved with a separate, more lightweight and specific metadata object (perhaps `FeatureMetadata`?). I'm not opposed to this.

> And the original thought was to keep params non related to feature calculation in the metadata

Sorry, I don't fully understand -- description is not related to feature calculation, so it would belong in metadata, no? Or do you mean that more generally you'd prefer not adding fields to the metadata?

pengyu-hou: > description is not related to feature calculation, so it would belong in metadata
Yes, correct. 
So we can think of a way to add the feature description to the existing metadata, rather than adding a new metadata field or feature metadata in the aggregation itself?

earangol-stripe: Do you mean something like how custom tags are handled? Basically keeping the descriptions in `Join` metadata through the custom JSON:

```python
def Derivation(name, expression, description):
   derivation = ttypes.Derivation(name, expression)
   if description:
      derivation.description = description
   return derivation


def Join(...):
   ...
   derivation_descriptions = {}
   for derivation in derivations:
      if derivation.description:
         derivation_descriptions[derivation.name] = derivation.description
   if derivation_descriptions:
      custom_json[derivation_descriptions] = derivation_descriptions
   ...
```

I would still prefer keeping the metadata closer to the object it refers to (perhaps with a more lightweight metadata object), but I wouldn't be too opposed to this alternative approach if you like that direction better. Or let me knoe if you meant something else.

pengyu-hou: had a discussion with @hzding621 and @camweston-stripe today. Let's reuse the `metaData` struct actually to use it in the aggregation itself. 

During the versioning check for join job, let's unset the meta data field before calculating the semantic hash so it won't affect the current behavior.

earangol-stripe: Great, thanks for the input! Could you point me to the versioning check that you're referring to? Do you mean [this validator](https://github.com/airbnb/chronon/blob/main/api/py/ai/chronon/repo/validator.py)?

pengyu-hou: It is at here: https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/JoinBase.scala#L390

PR #976, opened 2025-04-21
Set refreshIntervalMillis=8ms for metadata's force update

pengyu-hou:
## Summary

We are seeing potentially thread pool exhaustion during forcing update of GB serving info's metadata. Set a refresh limit to reduce the potential load.  

Next steps: 
- we will monitor the effectiveness of this PR. If not, we will implement an exponential backoff version 

## Why / Goal

The PR will make the online fetching more robust.  

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

### DIFF
--- a/online/src/main/scala/ai/chronon/online/TTLCache.scala
+++ b/online/src/main/scala/ai/chronon/online/TTLCache.scala
@@ -93,5 +93,4 @@ class TTLCache[I, O](f: I => O,
   def apply(i: I): O = asyncUpdateOnExpiry(i, ttlMillis)
   // manually refresh entry with a lower interval
   def refresh(i: I): O = asyncUpdateOnExpiry(i, refreshIntervalMillis)
-  def force(i: I): O = asyncUpdateOnExpiry(i, refreshIntervalMillis)
 }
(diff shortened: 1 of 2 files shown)


online/src/main/scala/ai/chronon/online/TTLCache.scala
```diff
@@ -93,5 +93,5 @@ class TTLCache[I, O](f: I => O,
   def apply(i: I): O = asyncUpdateOnExpiry(i, ttlMillis)
   // manually refresh entry with a lower interval
   def refresh(i: I): O = asyncUpdateOnExpiry(i, refreshIntervalMillis)
-  def force(i: I): O = asyncUpdateOnExpiry(i, 0)
+  def force(i: I): O = asyncUpdateOnExpiry(i, refreshIntervalMillis)
```
hzding621, line 96: so refresh/force becomes the same? should we just remove force?

hzding621 (approved):

online/src/main/scala/ai/chronon/online/TTLCache.scala
pengyu-hou: Good point

yuli-han (approved):

PR #978, opened 2025-04-23
Fixing bazel py:api_test dependency paths on fresh build

abbywh:
## Summary
when following devnotes.md on a fresh machine, `bazel build //...` correctly builds, but the target `bazel test //api/py:api_test` will fail. This will fail for these reasons:

* The thrift packages will resolve to `chronon.api.thrift.ai.chronon.api.ttypes` until the "chronon/api/thrift" import is added. 
* After adding and rebuilding, the target succeeds the import, even if you test on an earlier commit. This leads me to believe the current set up is slightly non hermatic and relies on some setup in the sbt instructions for thrift/python path resolution. 
* There are a couple more path problems with sqlglot and the lineage folder, which are also resolved here.

## Why / Goal
Fresh containers should be able to run tests with only the bazel instructions. 

## Test Plan
- [N] Added Unit Tests
- [Y] Covered by existing CI
- [N] Integration tested

## Checklist
- [ N/A] Documentation update

### DIFF
--- a/.ijwb/.bazelproject
+++ b/.ijwb/.bazelproject
@@ -14,4 +14,3 @@ additional_languages:
   python
   scala
   java
-  python
--- a/.ijwb/.bazelproject
+++ b/.ijwb/.bazelproject
@@ -14,3 +14,4 @@ additional_languages:
   python
   scala
   java
+  python
--- a/api/py/test/sample/production/joins/sample_team/sample_join.group_by_of_group_by
+++ b/api/py/test/sample/production/joins/sample_team/sample_join.group_by_of_group_by
@@ -6,7 +6,7 @@
     "customJson": "{\"check_consistency\": false, \"lag\": 0, \"join_tags\": null, \"join_part_tags\": {}}",
     "dependencies": [
       "{\"name\": \"wait_for_sample_namespace.sample_team_sample_staging_query_v1_ds\", \"spec\": \"sample_namespace.sample_team_sample_staging_query_v1/ds={{ ds }}\", \"start\": \"2021-03-01\", \"end\": null}",
-      "{\"name\": \"wait_for_sample_namespace.sample_team_sample_group_by_group_by_require_backfill_ds\", \"spec\": \"sample_namespace.sample_team_sample_group_by_group_by_require_backfill/ds={{ ds }}\", \"start\": \"2021-04-09\", \"end\": null}"
+      "{\"name\": \"wait_for_sample_namespace.sample_team_sample_group_by_require_backfill_ds\", \"spec\": \"sample_namespace.sample_team_sample_group_by_require_backfill/ds={{ ds }}\", \"start\": \"2021-04-09\", \"end\": null}"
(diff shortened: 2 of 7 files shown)


piyush-zlai (approved):
LGTM lets also see if @krisnaru / @knarukulla has cycles to look

requirements.txt
```diff
@@ -2,5 +2,5 @@ click
 thrift==0.13
 pytest
 twine==6.1.0
-
+sqlglot
```
pengyu-hou, line 5: @xiaohui-sun do you have a specific `sqlglot` version?

.ijwb/.bazelproject
```diff
@@ -14,3 +14,4 @@ additional_languages:
   python
   scala
   java
+  python
```
pengyu-hou, line 17: There is already `python` at line 14. 
```suggestion
```

api/py/test/sample/production/joins/sample_team/sample_join.group_by_of_group_by
```diff
@@ -6,7 +6,7 @@
     "customJson": "{\"check_consistency\": false, \"lag\": 0, \"join_tags\": null, \"join_part_tags\": {}}",
     "dependencies": [
       "{\"name\": \"wait_for_sample_namespace.sample_team_sample_staging_query_v1_ds\", \"spec\": \"sample_namespace.sample_team_sample_staging_query_v1/ds={{ ds }}\", \"start\": \"2021-03-01\", \"end\": null}",
-      "{\"name\": \"wait_for_sample_namespace.sample_team_sample_group_by_group_by_require_backfill_ds\", \"spec\": \"sample_namespace.sample_team_sample_group_by_group_by_require_backfill/ds={{ ds }}\", \"start\": \"2021-04-09\", \"end\": null}"
+      "{\"name\": \"wait_for_sample_namespace.sample_team_sample_group_by_require_backfill_ds\", \"spec\": \"sample_namespace.sample_team_sample_group_by_require_backfill/ds={{ ds }}\", \"start\": \"2021-04-09\", \"end\": null}"
```
pengyu-hou, line 9: This looks correct to me actually.

pengyu-hou: probably we need to follow up with the python test

pengyu-hou (approved):
Thanks @abbywh for putting up the PR. It looks good to me overall with some minor comments. 

@xiaohui-sun , could you please check:
- sqlglot version

requirements.txt
krisnaru: pin the version

xiaohui-sun: We can use 25.17.0. Thanks.

abbywh: updated with the lock

krisnaru:
LGTM

pengyu-hou:
> LGTM

Thanks @krisnaru for taking a look!

PR #983, opened 2025-05-02
Use configured JDK for thrift gen (Bazel)

tswitzer-netflix:
## Summary
This uses the bazel tools JDK for thrift generation (Python + Java) rather than relying on these being available on the system path.

## Why / Goal
Doesn't require pre-installing a system-wide JDK for thrift gen.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested


nikhil-zlai (approved):
thanks for the build improvement!

pengyu-hou (approved):

pengyu-hou:
Thanks @tswitzer-netflix

PR #989, opened 2025-05-10
Tuning Spark Test Performance

abbywh:
## Summary
By lowering default parallelism and keeping the metastore in memory, a few key bottlenecks around df saving and metastore bootstrapping are avoided, for example `repartitionAndWrite` now happens in <100ms.

Also separated out the join for better parallelism (this will only matter on a large build server due to cpu bounds) but also more cohesive testing groups/readable test files. 

Lowered some partition counts on JoinTests that were much larger than other tests. 

## Why / Goal
Test the spark suite a bit faster/more efficiently

My local test performance shaved ~10 minutes (30->20 minutes on a large build server) and a minute or two per single test class. 

## Test Plan

- [ N/A ] Added Unit Tests
- [  N/A ] Covered by existing CI
- [  N/A ] Integration tested

## Checklist
- [ N/A] Documentation update


spark/src/main/scala/ai/chronon/spark/SparkSessionBuilder.scala
```diff
@@ -64,6 +64,16 @@ object SparkSessionBuilder {
...
+      // ── use an in‐memory Derby metastore rather than disk + BoneCP ──
+      .config("spark.hadoop.javax.jdo.option.ConnectionURL", "jdbc:derby:memory:metastore_db;create=true")
+      .config("spark.hadoop.javax.jdo.option.ConnectionDriverName", "org.apache.derby.jdbc.EmbeddedDriver")
+      .config("spark.hadoop.javax.jdo.option.ConnectionUserName", "APP")
+      .config("spark.hadoop.javax.jdo.option.ConnectionPassword", "APP")
+      .config("spark.ui.enabled", "false")
+      .config("spark.chronon.outputParallelismOverride", "2")
+      .config("spark.chronon.group_by.parallelism", "2")
+      .config("spark.sql.shuffle.partitions", "2")
+      .config("spark.default.parallelism", "2")
```
pengyu-hou, line 76: let's move it to the local spark session builder
```suggestion
```

spark/src/main/scala/ai/chronon/spark/SparkSessionBuilder.scala
```diff
@@ -98,6 +108,9 @@ object SparkSessionBuilder {
       // use all threads - or the tests will be slow
         .master("local[*]")
         .config("spark.kryo.registrationRequired", s"${localWarehouseLocation.isEmpty}")
+        .config("spark.testing", "true")
+        .config("spark.ui.enabled", false)
+        .config("spark.sql.adaptive.enabled", true)
```
pengyu-hou, line 113: This is the local spark session builder. 
```suggestion
      // ── use an in‐memory Derby metastore rather than disk + BoneCP ──
        .config("spark.hadoop.javax.jdo.option.ConnectionURL", "jdbc:derby:memory:metastore_db;create=true")
        .config("spark.hadoop.javax.jdo.option.ConnectionDriverName", "org.apache.derby.jdbc.EmbeddedDriver")
        .config("spark.hadoop.javax.jdo.option.ConnectionUserName", "APP")
        .config("spark.hadoop.javax.jdo.option.ConnectionPassword", "APP")
        .config("spark.ui.enabled", "false")
        .config("spark.chronon.outputParallelismOverride", "2")
        .config("spark.chronon.group_by.parallelism", "2")
        .config("spark.sql.shuffle.partitions", "2")
        .config("spark.default.parallelism", "2")
        .config("spark.testing", "true")
        .config("spark.ui.enabled", false)
        .config("spark.sql.adaptive.enabled", true)
```

spark/src/main/scala/ai/chronon/spark/SparkSessionBuilder.scala
```diff
@@ -118,6 +131,7 @@ object SparkSessionBuilder {
     val userName = Properties.userName
     val baseBuilder = SparkSession
       .builder()
+      .config("spark.ui.enabled", "false")
```
pengyu-hou, line 134: removed as it will affect the prod spark streaming UI
```suggestion
```

pengyu-hou (commented):
Thanks for the contribution!

pengyu-hou:
@hzding621 how do you like the approach to split the join tests? It looks good to me.

spark/src/main/scala/ai/chronon/spark/SparkSessionBuilder.scala
```diff
@@ -98,6 +98,19 @@ object SparkSessionBuilder {
...
+        .config("spark.chronon.outputParallelismOverride", "2")
+        .config("spark.chronon.group_by.parallelism", "2")
+        .config("spark.sql.shuffle.partitions", "2")
+        .config("spark.default.parallelism", "2")
+        .config("spark.testing", "true")
+        .config("spark.ui.enabled", false)
+        .config("spark.sql.adaptive.enabled", true)
         .config("spark.local.dir", s"/tmp/$userName/$name")
         .config("spark.sql.warehouse.dir", s"$warehouseDir/data")
         .config("spark.hadoop.javax.jdo.option.ConnectionURL", metastoreDb)
```
tswitzer-netflix, line 116: Does the metastoreDb here need to match the in-memory one above?

abbywh: +1 Will change

nikhil-zlai (approved):
very cool!!

spark/src/main/scala/ai/chronon/spark/SparkSessionBuilder.scala
pengyu-hou: @tswitzer-netflix Thanks!

abbywh:
> > @hzding621 how do you like the approach to split the join tests? It looks good to me.
> 
> i like the idea 👍 the current `JoinTest` is too bloated to navigate

There's probably a more clean way to break it up than what I chose but I figured it would be a fine start

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
```diff
@@ -916,22 +651,24 @@ class JoinTest {
...
+    val diffCount = diff.count() 
     val queriesBare =
       tableUtils.sql(s"SELECT item, ts, ds from $itemQueriesTable where ds >= '$start' and ds <= '$dayAndMonthBefore'")
     assertEquals(queriesBare.count(), computed.count())
-    if (diff.count() > 0) {
-      println(s"Diff count: ${diff.count()}")
-      println(s"diff result rows")
+    if (diffCount > 0) {
+      logger.debug(s"Diff count: ${diffCount}")
+      logger.debug(s"diff result rows")
```
pengyu-hou, line 664: let us use `warn` actually if the diff requires attention. it will be easier for debugging
```suggestion
      logger.warn(s"Diff count: ${diffCount}")
      logger.warn(s"diff result rows")
```

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
```diff
@@ -916,22 +651,24 @@ class JoinTest {
                                      | JOIN queries
                                      | ON queries.item <=> part.item AND queries.ts <=> part.ts AND queries.ds <=> part.ds
                                      |""".stripMargin)
-    expected.show()
-
+    if (logger.isDebugEnabled){
```
pengyu-hou, line 654: where do we set the `isDebugEnabled`?

spark/src/test/scala/ai/chronon/spark/test/JoinBasicTest.scala
```diff
@@ -0,0 +1,531 @@
...
+    logger.debug("showing query result")
+    expected.show()
+    logger.debug(
+      s"Left side count: ${spark.sql(s"SELECT country, ds from $countryTable where ds >= '$start' and ds <= '$end'").count()}")
+    logger.debug(s"Actual count: ${computed.count()}")
+    logger.debug(s"Expected count: ${expected.count()}")
+    val diff = Comparison.sideBySide(computed, expected, List("country", "ds"))
+    if (diff.count() > 0) {
+      logger.debug(s"Diff count: ${diff.count()}")
+      logger.debug(s"diff result rows")
```
pengyu-hou, line 236:
```suggestion
      logger.warn(s"Diff count: ${diff.count()}")
      logger.warn(s"diff result rows")
```

spark/src/test/scala/ai/chronon/spark/test/JoinBasicTest.scala
```diff
@@ -0,0 +1,531 @@
...
+      logger.debug(
+        s"Left side count: ${spark.sql(s"SELECT user, ds from $namesTable where ds >= '$start' and ds <= '$end'").count()}")
+      logger.debug(s"Actual count: ${computed.count()}")
+      logger.debug(s"Expected count: ${expected.count()}")
+    }
+    val diff = Comparison.sideBySide(computed, expected, List("user", "ds"))
+    val diffCount = diff.count()
+    if (diffCount > 0) {
+      logger.debug(s"Diff count: $diffCount")
+      logger.debug(s"diff result rows")
```
pengyu-hou, line 525:
```suggestion
      logger.warn(s"Diff count: $diffCount")
      logger.warn(s"diff result rows")
```

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
```diff
@@ -514,8 +332,8 @@ class JoinTest {
     val diff = Comparison.sideBySide(computed, expected, List("item", "ts", "ds"))
 
     if (diff.count() > 0) {
-      println(s"Diff count: ${diff.count()}")
-      println(s"diff result rows")
+      logger.debug(s"Diff count: ${diff.count()}")
+      logger.debug(s"diff result rows")
```
pengyu-hou, line 336:
```suggestion
      logger.warn(s"Diff count: ${diff.count()}")
      logger.warn(s"diff result rows")
```

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
```diff
@@ -592,8 +410,8 @@ class JoinTest {
       tableUtils.sql(s"SELECT item, ts, ds from $itemQueriesTable where ds >= '$start' and ds <= '$dayAndMonthBefore'")
     assertEquals(queriesBare.count(), computed.count())
     if (diff.count() > 0) {
-      println(s"Diff count: ${diff.count()}")
-      println(s"diff result rows")
+      logger.debug(s"Diff count: ${diff.count()}")
+      logger.debug(s"diff result rows")
```
pengyu-hou, line 414:
```suggestion
      logger.warn(s"Diff count: ${diff.count()}")
      logger.warn(s"diff result rows")
```

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
```diff
@@ -658,8 +478,8 @@ class JoinTest {
...
```
pengyu-hou, line 482:
```suggestion
      logger.warn(s"Diff count: ${diff.count()}")
      logger.warn(s"diff result rows")
```

pengyu-hou (approved):
One last minor comment to use `warn` when diff count is larger than 0. Otherwise, LGMT. Thanks for the contribution.

spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala
abbywh: https://stackoverflow.com/questions/963492/in-log4j-does-checking-isdebugenabled-before-logging-improve-performance just to block the computation if debug is not set, we can change this to warn as well

abbywh:
@pengyu-hou I'll update isDebugEnabled to isWarnedEnabled shortly and it should be ready after that

pengyu-hou:
> isDebugEnabled

@abbywh Thanks so much. Maybe we can just keep `logger.warn` without `isWarnEnabled`? Or do you think if we can skip certain statements to speed up?

abbywh:
@pengyu-hou Yeah, I don't have the benchmarks (should have persisted them somewhere) but iirc it decently quick (a few seconds) but takes up a good bit of CPU power. I'm not opposed to cutting that part if you want, the spark tuning is the more significant part, but I'm not sure I understand the advantage.

pengyu-hou:
> @pengyu-hou Yeah, I don't have the benchmarks (should have persisted them somewhere) but iirc it decently quick (a few seconds) but takes up a good bit of CPU power. I'm not opposed to cutting that part if you want, the spark tuning is the more significant part, but I'm not sure I understand the advantage.

Ah I see what you mean. The string construction takes a good bit of CPU power. It is mostly used for debugging purpose if anything goes wrong in local or CI. But I guess it will improve a lot in CI because our CI is indeed consuming lots of CPU power. For local debugging process, we can iterate and see if we need to adjust it.           

This PR is good enough. Let's merge it! Thanks a lot!

PR #991, opened 2025-05-14
[WIP] Get bazel tests working

tswitzer-netflix:
## Summary
This gets tests running successfully under bazel.

WIP: Currently only with --define spark_version=3.2

## Why / Goal
They were not running and broken when I enabled them.

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update


nikhil-zlai:
if it is useful to you - we have a codemod script - that we used to re-write the tests into AnyFlatSpec - that work a lot better with bazel

https://github.com/zipline-ai/chronon/blob/main/scripts/codemod/test_replace.py

PR #992, opened 2025-05-15
fetcher refactor + support derivations in chaining

hzding621:
## Summary

Refactor fetch join calls and support derivations in chaining. 

- Refactor `fetchJoin` into multiple sub-functions: `fetchBaseJoin`, `fetchDerivations`, `fetchModelTransforms`  and `instrumentAndLog` 
- Update the call of `fetchJoin` from `JoinSourceRunner` into calling each sub-functions separately
  - for derivations, instead of calling `fetchDerivations`, use spark session to apply derivations 
- Also, support UDFs loading (via setups) in JoinSourceRunner

## Why / Goal

Currently, if a join contains derivations, it cannot be used as a joinSource for chaining, because in the chaining streaming job (`JoinSourceRunner`), spark executor will call `fetcher.fetchJoin` which invokes its own local spark session, and Spark doesn't like this - results in ```Caused by: org.apache.spark.SparkException: Failed to get broadcast_1_piece0 of broadcast_1```

Before model transforms chaining this is fine, because the same derivations can be expressed using GroupBy's query. But with model transforms chaining, running derivations as part of the join is a mandatory step to prepare model inputs. This is to be done together with UDF usage in derivations. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update


online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -288,101 +287,131 @@ class Fetcher(val kvStore: KVStore,
...
 
-  private def encode(loggingContext: Option[Metrics.Context],
+  private def doFetchJoin(requests: scala.collection.Seq[Request],
+                          joinConf: Option[api.Join] = None): Future[scala.collection.Seq[Response]] = {
+
+    val requestStartTs = System.currentTimeMillis()
+    val baseValuesF = fetchBaseJoin(requests, joinConf, requestStartTs)
+    val derivedValuesF = fetchDerivations(baseValuesF)
+    val modelTransformsF = fetchModelTransforms(derivedValuesF)
+    instrumentAndLog(modelTransformsF)
```
pengyu-hou: The flow is very clear. 

However, if we can simplify the how `requestStartTs` is used for instrumentation, it will be better. 

It works now and it is not related to this PR.

spark/src/main/scala/ai/chronon/spark/streaming/JoinSourceRunner.scala
```diff
@@ -416,38 +450,114 @@ class JoinSourceRunner(groupByConf: api.GroupBy, conf: Map[String, String] = Map
...
                      |ts: ${response.request.atMillis}
                      |request payload: ${response.request.keys}
-                     |response payload: ${response.values}
+                     |base response payload: ${response.baseValues}
+                     |""".stripMargin
+            }
+            logger.info(logMessage)
+          }
+
+          // convert to spark rows, appended with fetch start ts
```
pengyu-hou, line 457: minor refactor request: 
- add comments like step 1, convert to spark rows. step 2, call apply derivation. step 3, call model transform 
- Or, refactor into different methods so the flow is clearer.

pengyu-hou: Also, these three steps seem most computation intensive. Let's add more metrics to future optimization opportunities.

hzding621: All great ideas! will update

online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: Yes this indeeed can be simplified

PR #994, opened 2025-05-16
parallel batching of large join requests

hzding621:
## Summary

reduce batch size for really large batched join requests. there are sequential processing in fetchJoin kvstoreFuture.map{...} which becomes the bottleneck when the batch size is really large. 

## Why / Goal

latency reduction 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

integration testing:
- observed p99 latency reduction 600ms => 200ms (traffic: batch size ~150, ~17 join parts)

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @airbnb/airbnb-chronon-maintainers

### DIFF
--- a/online/src/main/scala/ai/chronon/online/Fetcher.scala
+++ b/online/src/main/scala/ai/chronon/online/Fetcher.scala
@@ -45,7 +45,6 @@ import scala.collection.mutable.ListBuffer
 import scala.collection.{Seq, mutable}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Random, Success, Try}
-import scala.util.ScalaJavaConversions.IterableOps
 
 object Fetcher {
   case class Request(name: String,
@@ -239,13 +238,11 @@ class Fetcher(val kvStore: KVStore,
 
     // Split requests into batches of size 32
     val batchSize = 32 // make tunable
+
     val batches = requests.grouped(batchSize).toSeq
-    batches.parallel.map(doFetchJoin(_, joinConf)).reduceLeft { (acc, batch) =>
-      acc.zip(batch).map {
-        case (accResponse, batchResponse) =>
-          accResponse ++ batchResponse
-      }
-    }
+    val batchFutures: Seq[Future[Seq[Response]]] =
+      batches.map(batch => doFetchJoin(batch, joinConf))
+    Future.sequence(batchFutures).map(_.flatten)
   }
 
   private def doFetchJoin(requests: scala.collection.Seq[Request],
--- a/online/src/main/scala/ai/chronon/online/Fetcher.scala
+++ b/online/src/main/scala/ai/chronon/online/Fetcher.scala
@@ -45,7 +45,6 @@ import scala.collection.mutable.ListBuffer
 import scala.collection.{Seq, mutable}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Random, Success, Try}
-import scala.util.ScalaJavaConversions.IterableOps
 
 object Fetcher {
   case class Request(name: String,
@@ -95,14 +94,16 @@ class Fetcher(val kvStore: KVStore,
               callerName: String = null,
               flagStore: FlagStore = null,
               disableErrorThrows: Boolean = false,
-              executionContextOverride: ExecutionContext = null)
+              executionContextOverride: ExecutionContext = null,
+              parallelBatchSize: Int = 32)
     extends FetcherBase(kvStore,
                         metaDataSet,
                         timeoutMillis,
                         debug,
                         flagStore,
                         disableErrorThrows,
-                        executionContextOverride) {
+                        executionContextOverride,
+                        parallelBatchSize) {
   private def reportCallerNameFetcherVersion(): Unit = {
     val message = s"CallerName: ${Option(callerName).getOrElse("N/A")}, FetcherVersion: ${BuildInfo.version}"
     val ctx = Metrics.Context(Environment.Fetcher)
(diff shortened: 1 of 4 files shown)


online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -234,6 +234,18 @@ class Fetcher(val kvStore: KVStore,
   }
 
   override def fetchJoin(requests: scala.collection.Seq[Request],
```
nikhil-zlai, line 236: the behavior imo suddenly becomes backwards incompatible - shall we introduce a new method `fetchJoinChunked(join, keys, chunkSize)` instead?

hzding621: i'm planning to make it a configurable parameter at Fetcher level.

hzding621: ~@nikhil-zlai why not make the chunking behavior the new default?~

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -94,7 +94,8 @@ class Fetcher(val kvStore: KVStore,
               callerName: String = null,
               flagStore: FlagStore = null,
               disableErrorThrows: Boolean = false,
-              executionContextOverride: ExecutionContext = null)
+              executionContextOverride: ExecutionContext = null,
+              joinFetchParallelChunkSize: Option[Int] = Some(32))
```
piyush-zlai: if we are going to keep this and not split like Nikhil mentioned, it might be good to keep the default behavior as before (unchunked) and allow clients to override if they want - wdyt?

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -234,7 +235,22 @@ class Fetcher(val kvStore: KVStore,
...
   override def fetchJoin(requests: scala.collection.Seq[Request],
-                         joinConf: Option[api.Join] = None): Future[scala.collection.Seq[Response]] = {
+                         joinConf: Option[Join] = None): Future[scala.collection.Seq[Response]] = {
+
+    if (joinFetchParallelChunkSize.isDefined) {
+      // If a chunk size is defined, split the requests into batches and process them in parallel
+      val batches = requests.grouped(joinFetchParallelChunkSize.get).toSeq
+      val batchFutures: Seq[Future[Seq[Response]]] =
+        batches.map(batch => doFetchJoin(batch, joinConf))
+      Future.sequence(batchFutures).map(_.flatten)
```
piyush-zlai, line 243: an idea here for the chunking might be to return individual chunked Futures rather than sequencing. With sequencing you are having to wait for all to get back, with individual futures you are getting a behavior akin to streaming responses back so the caller can start processing / acting on them

online/src/main/scala/ai/chronon/online/FetcherBase.scala
```diff
@@ -45,7 +45,8 @@ class FetcherBase(kvStore: KVStore,
                   debug: Boolean = false,
                   flagStore: FlagStore = null,
                   disableErrorThrows: Boolean = false,
-                  executionContextOverride: ExecutionContext = null)
+                  executionContextOverride: ExecutionContext = null,
+                  joinFetchParallelChunkSize: Option[Int] = Some(32))
```
pengyu-hou: should we use something like `Option(System.getProperty("ai.chronon.fetcher. join_fetch_parallel_chunk_size"))` to pass the default chunk size?

online/src/main/scala/ai/chronon/online/Fetcher.scala
pengyu-hou: Valid point. This is what GPT said: 
```
val batches = requests.grouped(joinFetchParallelChunkSize.get).toSeq

// Return individual futures instead of a single combined future
val responseFutures: Seq[Future[Seq[Response]]] = batches.map { batch =>
  doFetchJoin(batch, joinConf)
}

// Don't use Future.sequence - instead, process each future individually
responseFutures.foreach { future =>
  future.foreach { responses =>
    // Process each batch of responses as soon as it arrives
    processResponses(responses)  // Your processing function
  }
}
```

hzding621: @piyush-zlai @pengyu-hou thanks for the suggestion! updated and added a separate entry point for a chunked api in both scala and java fetcher

online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: Good point. updated!

online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: updated to keep existing behavior, ptal

online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: so I included two flows:
- existing fetchJoin will use Fetcher-level configuration (default to un-chunked). this is useful for clients who'd like to mostly keep existing behavior, but still able to enable chunking (but only at global level). I think Airbnb users will most likely use this flow. 
- new fetchJoinChunked for users who'd like to have fine-grained control. this API returns a `Seq<Future<Seq<Response>>>` for streaming-like behavior.

pengyu-hou (approved):

yuli-han (approved):

Issue #984, opened 2025-05-03
CHIP : Incremental batch feature aggregation


kambstreat:
# Problem Statement

Chronon reads data partitions based on window sizes to compute features for event sources. For example

* sum_clicks_past_30days          ←  30 day window
* count_visits_past_90days         ←  90 day window

When computing aggregations, one of the major time taking tasks is reading data from disk. If a feature needs 90 days lookback, Chronon reads 90 days event data to compute.

<img width="1103" alt="Image" src="https://github.com/user-attachments/assets/b9f61836-a089-411d-9d02-0125d4dc0d90" />

This CHIP is to add support for incremental aggregation which will avoid reading full dataset everytime. 

# Proposed Solution

Divide the aggregate computation into 2 parts.
* **Daily Aggregation** - Includes reading one day partition and store partial aggregates
* **Final Aggregation** - Read all partial aggregates and compute the final aggregates

For the above 2 steps to compute we assume

## Daily Aggregation IRs

Given a data source
* reads the latest partition and generates aggregated values (IRs)  
* These values will be stored in a new intermediate table.
* Schema of intermediate table is of IRs (different from final groupby output)

<img width="1115" alt="Image" src="https://github.com/user-attachments/assets/034c34b2-166b-4c62-9c53-f8179c51605c" />

The above diagram shows the daily IRs for `count` aggregation

## Final Aggregation (Final feature values)

* Read the intermediate table and compute the final aggregates
* Number of partitions read depends on the window size

## Pros
* Daily aggregates reduces the amnount of data read from disk
* Final aggregation is done on the intermediate table which is smaller than the original data source.

## Cons
* Intermediate table is created for each groupby which increases the storage cost
* Any changes in datasource will require re-computation of intermediate table.

# Optimizations

The proposed solution is a good start but can be improved further for invertible operators. Invertible operators are those which can be deleted. For example, `count` and `sum` are invertible operators.

### Requirements

* Previous day's groupBy output.   
  * groupBy values are basically ML features. So will refer this as `feature store`
* User activity table with latest and historical partitions.   

#### Case 1 : Unwindowed deletable

> Eg:  how many listings purchased by the user since signup.

These features are not windowed, which means the feature calculation happens over the lifetime.

<img width="596" alt="Image" src="https://github.com/user-attachments/assets/a5f1b65a-2fc1-448e-979f-3e35361b0e63" />

#### Case 2 : Windowed deletable
These features are windowed and has a inverse operator

[...]

PR #998, opened 2025-06-03
Write daily IRs for incremental aggregation

kambstreat:
## Summary

This PR address the following CHIP : https://github.com/airbnb/chronon/issues/984

## Why / Goal

The goal of the PR is to reduce the time taken to compute aggregations. Added daily level partial aggregates for batch feature computation, which avoids reading past day's event level data. 

With this change, we see, 4x improvement for computing features. (except first time run ) 
## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update


api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -176,8 +177,13 @@ object Extensions {
 
     def outputColumnName =
       s"${aggregationPart.inputColumn}_$opSuffix${aggregationPart.window.suffix}${bucketSuffix}"
+
+    def incOutputColumnName =
+      s"${aggregationPart.inputColumn}_$opSuffix${bucketSuffix}"
```
pengyu-hou, line 183: should we still keep the `aggregationPart.window.suffix`? Otherwise, how do we reconstruct the final output column?

aggregator/src/main/scala/ai/chronon/aggregator/row/RowAggregator.scala
```diff
@@ -70,6 +70,11 @@ class RowAggregator(val inputSchema: Seq[(String, DataType)], val aggregationPar
     .toArray
     .zip(columnAggregators.map(_.irType))
 
+  val incSchema = aggregationParts
```
pengyu-hou: I am thinking that we should use a full `incremental` in the code and we can keep `inc` as the suffix for the table so the table names are not getting too long. What do you think? 
```suggestion
  val incrementalSchema = aggregationParts
```

kambstreat: yes. good to use full name.

api/src/main/scala/ai/chronon/api/Extensions.scala
kambstreat: @pengyu-hou not sure I get it. I can not use the window.suffix right as the intermediate incremental is daily aggregation.

spark/src/test/scala/ai/chronon/spark/test/GroupByTest.scala
```diff
@@ -694,4 +696,96 @@ class GroupByTest {
...
              additionalAgg = aggs)
   }
+
+  @Test
+  def testIncrementalMode(): Unit = {
+    lazy val spark: SparkSession = SparkSessionBuilder.build("GroupByTest" + "_" + Random.alphanumeric.take(6).mkString, local = true)
+    implicit val tableUtils = TableUtils(spark)
+
+    val namespace = "test_incremental_group_by"
+
```
pengyu-hou, line 825:
```suggestion
tableUtils.createDatabase(namespace)
```

api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -98,6 +98,7 @@ object Extensions {
     def cleanName: String = metaData.name.sanitize
 
     def outputTable = s"${metaData.outputNamespace}.${metaData.cleanName}"
+    def incrementalOutputTable = s"${metaData.outputNamespace}.${metaData.cleanName}_inc"
```
pengyu-hou:
```suggestion
    def incrementalOutputTable = s"${metaData.outputNamespace}.${metaData.cleanName}_daily_inc"
```

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -117,6 +117,57 @@ class UniqueCount[T](inputType: DataType) extends SimpleAggregator[T, util.HashS
   }
 }
 
+class AverageIR extends SimpleAggregator[Array[Any], Array[Any], Double] {
```
abbywh: It looks to me like this mostly exists to serialize aggregators. Is there something more generic we can do? @nikhilsimha and I talked about https://fory.apache.org/ for example, but it could be even simpler to moving this logic into a different class abstraction

pengyu-hou: I agree that if we could somehow make it more general, then we would not have to implement for each operation.

pengyu-hou: also cc @nikhil-zlai 😁

pengyu-hou: I was thinking of sticking with Avro because it is what we used for the online serving path so that we can keep the ser//de logic same.

abbywh: I think the issue with Avro is that it needs to schematize the object, whereas hopIR is just an Array[Any]. If you just reflect the hopIR, there is no need to schematize it.

api/src/main/scala/ai/chronon/api/Extensions.scala
pengyu-hou: @abbywh I think we already have a daily here.

spark/src/main/scala/ai/chronon/spark/Driver.scala
```diff
@@ -467,7 +467,8 @@ object Driver {
         tableUtils,
         args.stepDays.toOption,
         args.startPartitionOverride.toOption,
-        !args.runFirstHole()
+        !args.runFirstHole(),
+        args.groupByConf.isIncremental
```
kambstreat: Add default value to the argument.

PR #1000, opened 2025-06-05
Passing app_name as argument to list_apps script

pallavia7:
## Summary
Passing app_name arguments helps in filtering output in the script itself to only return apps with the given name. 

## Why / Goal
We are migrating to new API the script uses to fetch applications and it takes in app_name now. So added this argument.

## Test Plan
Backward compatible, argument will be ignored if list_apps script is not updated
- [ ] Added Unit Tests
- [+ ] Covered by existing CI
- [ +] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/py/ai/chronon/repo/run.py
+++ b/api/py/ai/chronon/repo/run.py
@@ -419,7 +419,7 @@ class Runner:
                 print("Checking to see if a streaming job by the name {} already exists".format(self.app_name))
                 running_apps = []
                 try:
-                    running_apps = check_output("{}".format(self.list_apps_cmd)).decode("utf-8").split("\n")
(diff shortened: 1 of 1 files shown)


api/py/ai/chronon/repo/run.py
```diff
@@ -416,10 +453,24 @@ def run(self):
...
+                    )
+                )
                 running_apps = []
                 try:
-                    running_apps = check_output("{}".format(self.list_apps_cmd)).decode("utf-8").split("\n")
+                    running_apps = (
+                        check_output(
+                            "{} {} {}".format(
+                                self.list_apps_cmd,
+                                os.environ.get("EMR_CLUSTER"),
```
hzding621, line 467: we should generalize this as EMR is AWS specific. 

Maybe we can also pass in app_name via environment like APP_NAME=... . I think we already set APP_NAME in this way: https://github.com/airbnb/chronon/blob/main/api/py/ai/chronon/repo/run.py#L310-L326

PR #1001, opened 2025-06-08
Support aqe

abbywh:
## Summary
Testing out AQE over manual optimizing from the spark 2 era

## Why / Goal
See if we can get better performance with native spark 3 features to do things like coalesce and skew join

## Test Plan
- [ ] Added Unit Tests
- [ x ] Integration tested

Will benchmark myself and ask other teams to try it out


abbywh:
Spark tests down from 25->15 minutes, that's a 40% reduction 🤯

pengyu-hou:
> Spark tests down from 25->15 minutes, that's a 40% reduction 🤯

That is awesome!!

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -601,13 +601,24 @@ case class TableUtils(sparkSession: SparkSession) {
...
+      logger.info(
+        s"\n----[Running query with AQE auto coalesce enabled]----\n$query\n----[End of Query]----\n\n Query call path (not an error stack trace): \n$stackTraceStringPretty \n\n --------")
+    }
     try {
       // Run the query
-      val df = sparkSession.sql(query).coalesce(partitionCount)
-      df
+      val df = sparkSession.sql(query)
+      // if aqe auto coalesce is disabled, apply manual coalesce
+      val finalDf = if (!sparkSession.conf.get("spark.sql.adaptive.coalescePartitions.enabled", "true").toBoolean) {
```
pengyu-hou: should we use a constant or a variable to save the result of `sparkSession.conf.get("spark.sql.adaptive.coalescePartitions.enabled", "true").toBoolean`?

pengyu-hou: looks like there is already `useAqeRoute`, we could use something similar to refactor

abbywh: yeah that's a good idea

abbywh: We could overwrite it for some testing too. I still want to validate that this does something useful in a production environment generally before more investment though

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -594,6 +594,7 @@ case class TableUtils(sparkSession: SparkSession) {
 
   def sql(query: String): DataFrame = {
     val partitionCount = sparkSession.sparkContext.getConf.getInt("spark.default.parallelism", 1000)
+    val autoCoalesceEnabled = sparkSession.conf.get("spark.sql.adaptive.coalescePartitions.enabled", "true").toBoolean
```
jasonzou0, line 597: shall we default this to false if it is not set to match to the existing default behavior?

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -603,13 +604,24 @@ case class TableUtils(sparkSession: SparkSession) {
       .filter(_.contains("chronon"))
       .map(_.replace("at ai.chronon.spark.", ""))
       .mkString("\n")
-
-    logger.info(
-      s"\n----[Running query coalesced into at most $partitionCount partitions]----\n$query\n----[End of Query]----\n\n Query call path (not an error stack trace): \n$stackTraceStringPretty \n\n --------")
+    if (!autoCoalesceEnabled) {
```
jasonzou0, line 607: [nit] swap the if else branches to do the if enabled case first; i find code easier to read without explicit negations:)

Same with the `val finalDf` if-else below.

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -702,76 +714,91 @@ case class TableUtils(sparkSession: SparkSession) {
...
-          stats.get.count -> stats.get.partitionRange.partitions.length
-        } else {
-          val result = df.select(count(lit(1)), approx_count_distinct(col(partitionCol))).head()
-          (result.getAs[Long](0), result.getAs[Long](1).toInt)
-        }
-      } else {
-        (df.count(), 1)
-      }
+    val useAqeRoute = sparkSession.conf.getOption("spark.sql.adaptive.enabled").contains("true") &&
+      sparkSession.conf.getOption("spark.sql.adaptive.coalescePartitions.enabled").contains("true")
```
jasonzou0, line 718: Is it intentional "is AQE on" logic here check for both config options while the logic in `def sql()` above only checks for `spark.sql.adaptive.coalescePartitions.enabled`? 

It also might be a good idea to have a common helper function to abstract away the "is AQE on" logic.

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -702,76 +714,91 @@ case class TableUtils(sparkSession: SparkSession) {
                                           stats: Option[DfStats],
                                           sortByCols: Seq[String] = Seq.empty,
                                           partitionCols: Seq[String] = Seq.empty): Unit = {
-    // get row count and table partition count statistics
```
jasonzou0, line 705: is it possible to restructure this code to minimize diff and thus reviewer mental load to keep track of things? Can we do something like :

```
if (useAqeRoute) {
   write_into_df_the_simple_way()
   return
}

The_rest_of_the_non_aqe_code_as_is()
```

So the The_rest_of_the_non_aqe_code_as_is() code still remains on the same indentation level as is and wouldn't show up as diff in the PR.

spark/src/main/scala/ai/chronon/spark/ChrononKryoRegistrator.scala
```diff
@@ -147,7 +147,11 @@ class ChrononKryoRegistrator extends KryoRegistrator {
       "org.apache.spark.sql.catalyst.InternalRow$$anonfun$getAccessor$8",
       "org.apache.spark.sql.catalyst.InternalRow$$anonfun$getAccessor$5",
       "scala.collection.immutable.ArraySeq$ofRef",
-      "org.apache.spark.sql.catalyst.expressions.GenericInternalRow"
+      "org.apache.spark.sql.catalyst.expressions.GenericInternalRow",
```
jasonzou0, line 150: add an inline comment that these registrations are needed for AQE (that's just my guess on why they are here:))?

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -702,76 +714,91 @@ case class TableUtils(sparkSession: SparkSession) {
...
+        logger.info(s"Input DataFrame for table $tableName is empty. Nothing to write.")
+        return
       }
-      val dailyFileCount = outputParallelism.getOrElse(dailyFileCountBounded)
-
-      // finalized shuffle parallelism
-      val shuffleParallelism = Math.max(dailyFileCount * nonZeroTablePartitionCount, minWriteShuffleParallelism)
-      val saltCol = "random_partition_salt"
-      val saltedDf = df.withColumn(saltCol, round(rand() * (dailyFileCount + 1)))
+      val sortedDf = df.sortWithinPartitions(sortByCols.map(col).toSeq: _*)
```
jasonzou0, line 725: why are we ignoring the `partitionCols` that users may pass in?

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
abbywh: I think the idea is to change the default behavior since it's much more performant in most cases

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
abbywh: This is a common style in this code base, ex: `if (!tableExists(tableName)) return Seq.empty[String]` which I chose to follow (I agree it's a bit wonky)

spark/src/main/scala/ai/chronon/spark/ChrononKryoRegistrator.scala
abbywh: It's actually more generic than that, AQE uses it, but some of the SQL expressions will output this too. I think this matches the current file organization.

A separate PR/github issues could be cloning this and adding more explanations to the class
https://github.com/apache/spark/blob/dc687d4c83b877e90c8dc03fb88f13440d4ae911/core/src/main/scala/org/apache/spark/serializer/KryoSerializer.scala#L575

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
abbywh: Good question, this is kind of a weird one, but iirc, and it's been awhile, including them actually *breaks* the table schema assumptions and fails the test. 

We are essentially relying on AQE as well as the writer plugin to do our repartitioning into the correct partitions. The catalyst engine rewrites this entire section to do the repartition before the sort iirc, making us know the partition columns are all equal to each other, and therefore we don't sort them here. Including it changes the plan in such a way that it does not match the previous chronon plan. Maybe this is a catalyst bug or something, but I'm pretty sure I ran into the same issue as stripe where the results would sometimes be out of order until I removed that.

spark/src/main/scala/ai/chronon/spark/TableUtils.scala
abbywh: We eventually branch into the same code below, not sure how to make this cleaner while keeping that.  I definitely emphasize that this diff rendered weirdly

PR #1003, opened 2025-06-14
Update get table properties function for iceberg table

yuli-han:
## Summary

The sparkSession.sessionState.catalog.getTempViewOrPermanentTableMetadata function doesn't work for iceberg tables. 
An exception would be thrown 
```
java.lang.ClassNotFoundException: Class org.apache.iceberg.mr.hive.HiveIcebergSerDe not found
        at org.apache.hadoop.conf.Configuration.getClassByName(Configuration.java:2273)
```
The PR is to use standard [SQL DDL](https://spark.apache.org/docs/latest/sql-ref-syntax-aux-show-tblproperties.html) to get table properties for iceberg tables. 

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
+++ b/spark/src/main/scala/ai/chronon/spark/TableUtils.scala
@@ -916,8 +916,18 @@ case class TableUtils(sparkSession: SparkSession) {
 
   def getTableProperties(tableName: String): Option[Map[String, String]] = {
     try {
-      val tableId = sparkSession.sessionState.sqlParser.parseTableIdentifier(tableName)
-      Some(sparkSession.sessionState.catalog.getTempViewOrPermanentTableMetadata(tableId).properties)
+      tableFormatProvider.readFormat(tableName) match {
+        case Iceberg =>
+          val props = sparkSession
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/ai/chronon/spark/TableUtils.scala
```diff
@@ -916,8 +916,18 @@ case class TableUtils(sparkSession: SparkSession) {
...
-      Some(sparkSession.sessionState.catalog.getTempViewOrPermanentTableMetadata(tableId).properties)
+      tableFormatProvider.readFormat(tableName) match {
+        case Iceberg =>
+          val props = sparkSession
+            .sql(s"SHOW TBLPROPERTIES $tableName")
+            .collect()
+            .map(row => row.getString(0) -> row.getString(1))
+            .toMap
+          Some(props)
+        case _ =>
```
pengyu-hou, line 927: let's add a comment the default case is Hive. 
Or we can have three cases: 
- ice berg
- hive
- default: not supported format yet.

PR #1005, opened 2025-06-25
Pre-aggregated Entities fix when time column was set

pengyu-hou:
## Summary

Quote from @camweston-stripe and thanks @camweston-stripe  for reporting the issue.

I think I found a bug in the Chronon Spark code for GroupByUpload that affects Snapshot Entity Group Bys that don't have any aggregations. The problem occurs when a time column is specified—causing an off-by-one misalignment between the schema and data. Time columns are not required/utilized in the Snapshot Entity no agg case, but from what I understand, there is nothing in place to stop the user from providing one in their source.
Can I get another set of eyes on the info below to confirm my analysis/suggested fix is sound?
### Issue:
One of our GBU jobs was failing with java.lang.ClassCastException errors. The problem was caused by inconsistent column filtering: the column defined as Constants.TimeColumn (e.g., _internal_time_column or ts) was omitted in the schema but included in the data array. This created an off-by-one alignment error between the schema and data.
### Impact:
When Chronon’s conversion in [ai.chronon.api.Row.to](https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Row.scala#L128) processed fields—expecting a specific type but receiving data from the wrong position—a type mismatch occurred. This resulted in repeated task failures and job abortion. Our concern here is that it's possible for this to fail silently. For example, if the value at hand can be successfully casted to the misaligned type, then no error is thrown.
### Initial Error Analysis:
The job produced two ClassCastException errors:
Long cannot be cast to Boolean (20250607 data).
String cannot be cast to Number (20250608 data).
Both errors pointed to type mismatches in [ai.chronon.api.Row.to](https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Row.scala#L128).
### Source Configuration Adjustments & Symptom Shift:
Explicitly casting state to STRING resolved one error but made the other (“Long to Boolean”) consistent.
Additional CAST(... AS BOOLEAN) statements were introduced for boolean columns.
### PySpark Data Profiling:
Scripts confirmed that "boolean candidate" columns were already of BooleanType and that numeric columns did not consist solely of 0s and 1s, ruling out simple casting errors.
### Diagnostic Logging Enhancements:
Implemented log statements in [AvroConversions.encodeBytes](https://github.com/airbnb/chronon/blob/main/online/src/main/scala/ai/chronon/online/AvroConversions.scala#L171) and [Row.to](https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Row.scala#L128) (including recursive calls) to print the value, expected data type, current field, and target Avro type.
Logs revealed misalignment: for instance
```
EBUG_ROW_TO_ENTER: Value='NA' (type: java.lang.String), ChrononDataType(raw)='StringType', Field='region' //ordering is still correct at this point
DEBUG_ROW_TO_ENTER: Value='1749254400000' (type: java.lang.Long), ChrononDataType(raw)='LongType', Field='rejected_dupes_by_url' //off by one issue starts here (value is the ts, field is not)
DEBUG_ROW_TO_ENTER: Value='0' (type: java.lang.Long), ChrononDataType(raw)='LongType', Field='dupes_by_owner_name_dob' // long -> long silent failure (SF)
DEBUG_ROW_TO_ENTER: Value='1' (type: java.lang.Long), ChrononDataType(raw)='LongType', Field='dupes_by_device_id' //long -> long SF
DEBUG_ROW_TO_ENTER: Value='1' (type: java.lang.Long), ChrononDataType(raw)='BooleanType', Field='is_funded_by_top_investor' //long -> boolean SF
DEBUG_ROW_TO_ENTER: Value='false' (type: java.lang.Boolean), ChrononDataType(raw)='BooleanType', Field='is_growth' //bool --> bool SF
DEBUG_ROW_TO_ENTER: Value='false' (type: java.lang.Boolean), ChrononDataType(raw)='StringType', Field='month' //bool --> str SF
DEBUG_ROW_TO_ENTER: Value='month-6' (type: java.lang.String), ChrononDataType(raw)='LongType', Field='rejected_dupes_by_owner_name_dob' //str --> long loud failure
```
- This indicated that data from one column was being applied to the next.
### Pinpointing the Misalignment:
- The data array for [Row.to](https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Row.scala#L128) is built in the snapshotEntities method in [GroupByUpload.scala](https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/GroupByUpload.scala#L44).
- The array is created with [valuesIndices.map(row.get)](https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/GroupByUpload.scala#L50) based on a filtered list of column names.
- The filtering logic here differed from that in [GroupBy.scala](https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/GroupBy.scala#L69) (where Constants.TimeColumn is excluded), causing the off-by-one error.
3. In-depth Analysis of the Issue
The root cause was two different filtering approaches:
Schema Definition (preAggSchema in GroupBy.scala):
The schema is constructed by filtering out key and reserved columns (including Constants.TimeColumn), so the schema never expects the internal timestamp.
Data Array Construction (snapshotEntities in GroupByUpload.scala):
The filtering here only excludes key columns and the partition column; it misses excluding Constants.TimeColumn, meaning the corresponding value is included in the data array.
### Consequence:
Assume the input order is:
col_A, _internal_time_column, col_B
#### Data array order: col_A, _internal_time_column, col_B
This misalignment means that when processing col_B, the job mistakenly reads _internal_time_column (or later fields), causing type mismatches (e.g., expecting a Long but receiving a String).
## 4. The Fix
To resolve the issue, we should be adding filtering logic for both the schema and the data array to make sure they are aligned. In the corrected code for GroupByUpload.snapshotEntities, the filtering would now mirror that of GroupBy.scala:
```
// In GroupByUpload.scala
 def snapshotEntities: KvRdd = {
    if (groupBy.aggregations == null || groupBy.aggregations.isEmpty) {
      val keysAndPartitionForBuilder = (groupBy.keyColumns :+ tableUtils.partitionColumn).toArray
      val keyBuilder = FastHashing.generateKeyBuilder(keysAndPartitionForBuilder, groupBy.inputDf.schema)

      //Ensure that other meta columns like `_internal_time_column` are also removed.
      //We do this so that our key/values align properly in KvRDD
      val keysAndMetaColumns: Seq[String] = groupBy.keyColumns ++ Constants.ReservedColumns(tableUtils.partitionColumn)
      val dataArrayColumnNames: Seq[String] = groupBy.inputDf.schema.map(_.name).filterNot(keysAndMetaColumns.contains).sorted
      val preAggSchemaFieldNames: Seq[String] = groupBy.preAggSchema.fields.map(_.name).sorted

      assert(
        dataArrayColumnNames == preAggSchemaFieldNames,
        s"Misalignment! Data array columns: [${dataArrayColumnNames.mkString(", ")}] vs Pre-agg Schema fields: [${preAggSchemaFieldNames.mkString(", ")}]"
      )

      val valuesIndices = dataArrayColumnNames.map(groupBy.inputDf.schema.fieldIndex).toArray

      val rdd = groupBy.inputDf.rdd
        .map { row =>
          keyBuilder(row).data.init -> valuesIndices.map(row.get)
        }
      KvRdd(rdd, groupBy.keySchema, groupBy.preAggSchema)
    } else {
      fromBase(groupBy.snapshotEntitiesBase)
    }
  }
```
By filtering columns with groupBy.keyColumns ++ Constants.ReservedColumns(tableUtils.partitionColumn), the data array no longer includes Constants.TimeColumn, ensuring alignment with preAggSchema. This resolves the off-by-one error and stops the ClassCastException.

This PR is ported from @nikhil-zlai https://github.com/zipline-ai/chronon/pull/827 

## Why / Goal

It will correct the case when user specified time column in a no-agg groupBy

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@airbnb/airbnb-chronon-maintainers @camweston-stripe @nikhil-zlai @piyush-zlai @caiocamatta-stripe


spark/src/main/scala/ai/chronon/spark/FastHashing.scala
```diff
@@ -114,7 +115,14 @@ object FastHashing {
...
       }
       val hashCode = hasher.hash()
-      KeyWithHash(keyIndices.map(row.get), hashCode.asBytes(), hashCode.asInt())
+      val keys = Array.fill[Any](keyIndices.length)(null)
+      var i = 0
+      while (i < keyIndices.length) {
+        val index = keyIndices(i)
+        keys.update(i, row.get(index))
+        i += 1
+      }
```
pkundurthy, line 124: nit: more idiomatic scala using `zipWithIndex`
```
keyIndices.zipWithIndex.foreach { case (index, i) =>
  keys.update(i, row.get(index))
}
```

pkundurthy (approved):
LGTM

spark/src/main/scala/ai/chronon/spark/FastHashing.scala
hzding621: +1

pengyu-hou: Actually, this should be intended because native for/while loop performs better than the collection high-order function and uses less memory

PR #1015, opened 2025-07-17
StagingQuery create view and virtual partitions

pengyu-hou:
## Summary

This PR will introduce a param called `createView` for Staging Query. Instead of writing its output to data warehouse, it will create a view instead. This implementation provides virtual partition table support for Chronon staging query views, enabling efficient orchestration and dependency management for views that contain all historical data **without** date range filtering. The date range will be pushed down to the downstream tasks like a GroupBy job. 

Because view does not have regular partitions like materialized tables, this PR will create a signal partition metadata table, which can be later used in orchestration code to sense. The current implementation will create this table with `ds` and `table_name` as partition columns, therefore, for Airflow users, there will be no more Airflow code changes needed. 

```sql
CREATE TABLE chronon_signal_partitions (
  created_at STRING        -- Creation timestamp (data column)
) 
PARTITIONED BY (
  ds STRING,              -- Partition date (e.g., "2025-08-14")
  table_name STRING       -- Staging query view name
)
```

image

A `staging_query.py` Python wrapper is implemented for `StagingQueryEventSource` and `StagingQueryEntitySource`.

## Why / Goal

- Support creating a view for staging query without materializing the table. 
- Add a signal partition table for job orchestration 
- Add a staging event/entity source for explicit deps 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [x] tested on gateway machines 

```
python3 ./scripts/run.py --mode=backfill --conf=production/staging_queries/zipline_test/sample_staging_query_view.v1  --ds 2025-08-13 --chronon-jar ~/test/chronon-spark.jar | tee ~/test/view.log

python3 ./scripts/run.py --mode=backfill --conf=production/group_bys/zipline_test/test_view_source.v2  --ds 2025-08-12 --chronon-jar ~/test/chronon-spark.jar | tee ~/test/view_groupby.log
```
## Checklist
- [ ] Documentation update

## Reviewers

### DIFF
--- a/api/py/ai/chronon/utils.py
+++ b/api/py/ai/chronon/utils.py
@@ -353,7 +353,7 @@ def get_dependencies(
             )
         elif src.entities:
             table_name = src.entities.snapshotTable
-            
+
             if hasattr(src.entities, 'is_input_staging_query_view') and src.entities.is_input_staging_query_view:
                 source_namespace = getattr(src.entities, 'namespace', table_name.split('.')[0])
                 clean_table_name = table_name.split('.')[-1] if '.' in table_name else table_name
@@ -372,7 +372,7 @@ def get_dependencies(
             result = [wait_for_simple_schema(parentJoinOutputTable, lag, start, end, query=src.joinSource.query)]
         else:
             table_name = src.events.table
-            
+
             if hasattr(src.events, 'is_input_staging_query_view') and src.events.is_input_staging_query_view:
                 source_namespace = getattr(src.events, 'namespace', table_name.split('.')[0])
                 clean_table_name = table_name.split('.')[-1] if '.' in table_name else table_name
--- a/api/py/ai/chronon/utils.py
+++ b/api/py/ai/chronon/utils.py
@@ -299,7 +299,6 @@ def log_table_name(obj, full_name: bool = False):
 def get_staging_query_output_table_name(
     staging_query: api.StagingQuery, full_name: bool = False
 ):
-    """generate output table name for staging query job"""
     __set_name(staging_query, api.StagingQuery, "staging_queries")
     return output_table_name(staging_query, full_name=full_name)
 
@@ -353,16 +352,36 @@ def get_dependencies(
                 )
             )
         elif src.entities:
-            result = [
-                wait_for_simple_schema(src.entities.snapshotTable, lag, start, end, query=src.entities.query)
-            ]
+            table_name = src.entities.snapshotTable
+            
+            if hasattr(src.entities, 'is_input_staging_query_view') and src.entities.is_input_staging_query_view:
+                source_namespace = getattr(src.entities, 'namespace', table_name.split('.')[0])
+                clean_table_name = table_name.split('.')[-1] if '.' in table_name else table_name
+                result = [{
+                    "name": f"wait_for_view_{clean_table_name}_ds{'_minus_' + str(lag) if lag > 0 else ''}",
+                    "spec": f"{source_namespace}.chronon_virtual_partitions/ds={{{{ ds }}}}/table_name={table_name}"
+                }]
+            else:
+                result = [
+                    wait_for_simple_schema(table_name, lag, start, end, query=src.entities.query)
+                ]
         elif src.joinSource:
             parentJoinOutputTable = get_join_output_table_name(
                 src.joinSource.join, True
             )
             result = [wait_for_simple_schema(parentJoinOutputTable, lag, start, end, query=src.joinSource.query)]
         else:
-            result = [wait_for_simple_schema(src.events.table, lag, start, end, query=src.events.query)]
+            table_name = src.events.table
+            
+            if hasattr(src.events, 'is_input_staging_query_view') and src.events.is_input_staging_query_view:
+                source_namespace = getattr(src.events, 'namespace', table_name.split('.')[0])
+                clean_table_name = table_name.split('.')[-1] if '.' in table_name else table_name
+                result = [{
+                    "name": f"wait_for_view_{clean_table_name}_ds{'_minus_' + str(lag) if lag > 0 else ''}",
+                    "spec": f"{source_namespace}.chronon_virtual_partitions/ds={{{{ ds }}}}/table_name={table_name}"
+                }]
+            else:
+                result = [wait_for_simple_schema(table_name, lag, start, end, query=src.events.query)]
     return [json.dumps(res) for res in result]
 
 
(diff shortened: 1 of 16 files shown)


api/thrift/api.thrift
```diff
@@ -46,6 +46,11 @@ struct StagingQuery {
     * Spark SQL setup statements. Used typically to register UDFs.
     **/
     4: optional list<string> setups
+    /**
+    * If true, creates a view in the warehouse (for intermediate tables).
+    * If false, creates a table in the warehouse (for final tables).
+    **/
+    5: optional bool createView = false
```
hzding621: will `= false` break the backward compatibility check here https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/ThriftJsonCodec.scala#L84-L86? 


should we just do `5: optional bool createView`

pengyu-hou: Great catch!

spark/src/main/scala/ai/chronon/spark/StagingQuery.scala
```diff
@@ -46,7 +49,12 @@ class StagingQuery(stagingQueryConf: api.StagingQuery, endPartition: String, tab
     Option(stagingQueryConf.setups).foreach(_.toScala.foreach(tableUtils.sql))
     // the input table is not partitioned, usually for data testing or for kaggle demos
     if (stagingQueryConf.startPartition == null) {
-      tableUtils.sql(stagingQueryConf.query).save(outputTable)
+      if (Option(stagingQueryConf.createView).getOrElse(false)) {
```
hzding621: in `Driver.scala` it seems we already handle createView=true/false separately, and `computeStagingQuery` will only be called when `createView=false`, so here should we throw exception instead?

spark/src/main/scala/ai/chronon/spark/StagingQuery.scala
```diff
@@ -77,9 +85,16 @@ class StagingQuery(stagingQueryConf: api.StagingQuery, endPartition: String, tab
             val renderedQuery =
               StagingQuery.substitute(tableUtils, stagingQueryConf.query, range.start, range.end, endPartition)
             logger.info(s"Rendered Staging Query to run is:\n$renderedQuery")
-            val df = tableUtils.sql(renderedQuery)
-            tableUtils.insertPartitions(df, outputTable, tableProps, partitionCols, autoExpand = enableAutoExpand.get)
-            logger.info(s"Wrote to table $outputTable, into partitions: $range $progress")
+
+            if (Option(stagingQueryConf.createView).getOrElse(false)) {
```
hzding621: ditto: throw exception instead?

spark/src/test/scala/ai/chronon/spark/test/GroupByTest.scala
```diff
@@ -853,4 +853,108 @@ class GroupByTest {
...
+    assertEquals(0, diff.count())
+
+    // Verify we can query the results successfully with partition pushdown
+    val filteredResult = tableUtils.sql(s"""
+      SELECT user, time_spent_ms_count 
+      FROM $groupByOutputTable 
+      WHERE ds >= '${tableUtils.partitionSpec.minus(today, new Window(7, TimeUnit.DAYS))}'
+    """)
+    assertTrue("Should be able to filter GroupBy results", filteredResult.count() >= 0)
+  }
```
hzding621, line 959: very nice test!!!

spark/src/test/scala/ai/chronon/spark/test/StagingQueryTest.scala
```diff
@@ -293,4 +293,105 @@ class StagingQueryTest {
...
+    stagingQueryUnset.computeStagingQuery(stepDays = Option(30))
+
+    val outputUnset = stagingQueryConfUnset.metaData.outputTable
+    val isTableUnset = tableUtils.tableReadFormat(outputUnset) match {
+      case View => false
+      case _ => true
+    }
+    
+    assert(isTableUnset, s"Expected $outputUnset to be a table when createView is unset")
+  }
```
hzding621, line 396: 👍

api/py/ai/chronon/utils.py
```diff
@@ -352,16 +351,36 @@ def get_dependencies(
                 )
             )
         elif src.entities:
-            result = [
-                wait_for_simple_schema(src.entities.snapshotTable, lag, start, end, query=src.entities.query)
-            ]
+            table_name = src.entities.snapshotTable
+
+            if hasattr(src.entities, 'is_input_staging_query_view') and src.entities.is_input_staging_query_view:
```
hzding621: Maybe we can create Python wrappers like `StagingQueryEventSource` and `StagingQueryEntitySource` in the OSS repo. For StagingQuery-based sources, we can guarantee that there is no realtime component (it's always batch only). And the wrappers can take care of:
- getting the actual table (it can take in a StagingQuery python object)
- creating the dependencies for the virtual tables

spark/src/main/scala/ai/chronon/spark/StagingQuery.scala
pengyu-hou: ah good catch. I was making changes here first and forgot to clean it up.

pengyu-hou: I added assertion to the top of this method. It is cleaner.

api/py/ai/chronon/utils.py
```diff
@@ -299,7 +299,6 @@ def log_table_name(obj, full_name: bool = False):
 def get_staging_query_output_table_name(
     staging_query: api.StagingQuery, full_name: bool = False
 ):
-    """generate output table name for staging query job"""
```
hzding621: keep?

hzding621 (approved):

api/py/ai/chronon/utils.py
pengyu-hou: added it back!

spark/src/main/scala/ai/chronon/spark/StagingQuery.scala
```diff
@@ -29,6 +30,8 @@ class StagingQuery(stagingQueryConf: api.StagingQuery, endPartition: String, tab
   @transient lazy val logger = LoggerFactory.getLogger(getClass)
   assert(Option(stagingQueryConf.metaData.outputNamespace).nonEmpty, s"output namespace could not be empty or null")
   private val outputTable = stagingQueryConf.metaData.outputTable
+  val signalPartitionsTable =
+    s"${stagingQueryConf.metaData.outputNamespace}.${StagingQuery.SIGNAL_PARTITIONS_TABLE_NAME}"
```
hzding621, line 34: Are we not using a global signal table for all configs, but instead creating one for each team?

pengyu-hou: @hzding621 yes, the current approach use one signal table for each team. Or should we actually just use one table?

pkundurthy (approved):

PR #1016, opened 2025-07-18
Python wrappers for source objects

pyalex:
## Summary

Adding functions to the Python SDK for source objects creation. This is a fully backward-compatible change.

## Why / Goal
The primary motivation is to enable the addition of extra attributes at the source level. Similarly to how it's done in `GroupBy` and `Join`: all extra arguments are stored in the `customJson` attribute. Sources can have all sorts of metadata, ie `bootstrap.server` for Kafka source.

Additional benefits:
* Slightly less verbose API
```
my_source = ttypes.Source(
    events=ttypes.EventSource(
          table=...
    )
)
```
vs
```
my_source = source.EventSource(
    table=...
)
```
* Improving API consistency: existing Python wrappers (ie, `GroupBy`, `Join`) use Pythonic snake case for parameter names, whereas code generated from Thrift uses camel case (ie, `snapshotTable` in `EntitySource`)

## Test Plan

- [ ] Added Unit Tests
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/py/test/sample/production/group_bys/sample_team/event_sample_group_by.v1
+++ b/api/py/test/sample/production/group_bys/sample_team/event_sample_group_by.v1
@@ -25,7 +25,8 @@
           "startPartition": "2021-04-09",
           "timeColumn": "ts",
           "setups": []
-        }
+        },
+        "customJson": "{}"
       }
     }
   ],
--- a/api/thrift/api.thrift
+++ b/api/thrift/api.thrift
@@ -68,6 +68,11 @@ struct EventSource {
     * If each new hive partition contains not just the current day's events but the entire set of events since the begininng. The key property is that the events are not mutated across partitions.
     **/
     4: optional bool isCumulative
+
+    /**
+    * Any extra attributes can be stored here. Ie, Kafka bootstrap servers for a streaming source, or AWS IAM role for accessing Iceberg table
+    **/
+    5: optional string customJson
 }
 
 
@@ -98,6 +103,11 @@ struct EntitySource {
     * The logic used to scan both the table and the topic. Contains row level transformations and filtering expressed as Spark SQL statements.
     */
     4: optional Query query
+
+    /**
+    * Any extra attributes can be stored here. Ie, Kafka bootstrap servers for a streaming source, or AWS IAM role for accessing Iceberg table
+    **/
+    5: optional string customJson
 }
 
 struct ExternalSource {
--- a/api/py/test/lineage/test_parse_group_by.py
+++ b/api/py/test/lineage/test_parse_group_by.py
@@ -16,6 +16,7 @@ import os
 import unittest
 
 from ai.chronon import group_by
+from ai.chronon import source
 from ai.chronon.api import ttypes
 from ai.chronon.group_by import Accuracy, Derivation
 from ai.chronon.lineage.lineage_metadata import ConfigType, TableType
@@ -25,7 +26,7 @@ from helper import compare_lineages
 
 class TestParseGroupBy(unittest.TestCase):
     def setUp(self):
-        gb_event_source = ttypes.EventSource(
+        gb_event_source = source.EventSource(
             table="source.gb_table",
             topic=None,
             query=ttypes.Query(
@@ -35,7 +36,7 @@ class TestParseGroupBy(unittest.TestCase):
             ),
         )
 
-        gb_event_source1 = ttypes.EventSource(
+        gb_event_source1 = source.EventSource(
             table="source.gb_table1",
             topic=None,
             query=ttypes.Query(
--- a/api/py/test/lineage/test_parse_join.py
+++ b/api/py/test/lineage/test_parse_join.py
@@ -16,6 +16,7 @@ import os
 import unittest
 
 from ai.chronon import group_by
+from ai.chronon import source
 from ai.chronon.api import ttypes
 from ai.chronon.api import ttypes as api
 from ai.chronon.group_by import Derivation
@@ -27,7 +28,7 @@ from helper import compare_lineages
 
 class TestParseJoin(unittest.TestCase):
     def setUp(self):
-        gb_event_source = ttypes.EventSource(
+        gb_event_source = source.EventSource(
             table="gb_table",
             topic=None,
             query=ttypes.Query(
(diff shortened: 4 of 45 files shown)


api/py/ai/chronon/source.py
```diff
@@ -0,0 +1,81 @@
...
+    :return:
+      A source object of kind EventSource
+    """
+    return ttypes.Source(
+        events=ttypes.EventSource(
+            table=table,
+            topic=topic,
+            query=query,
+            isCumulative=is_cumulative,
+            customJson=json.dumps(kwargs),
```
nikhil-zlai: add a check to ensure that *kwargs doesn't contain a key - that is a param of eventSource.

api/py/ai/chronon/source.py
```diff
@@ -0,0 +1,81 @@
...
+    :return:
+      A source object of kind EntitySource
+    """
+    return ttypes.Source(
+        entities=ttypes.EntitySource(
+            snapshotTable=snapshot_table,
+            mutationTable=mutation_table,
+            query=query,
+            mutationTopic=mutation_topic,
+            customJson=json.dumps(kwargs),
```
nikhil-zlai: same as above.

api/py/test/sample/production/group_bys/sample_team/event_sample_group_by.v1
```diff
@@ -25,7 +25,8 @@
           "startPartition": "2021-04-09",
           "timeColumn": "ts",
           "setups": []
-        }
+        },
+        "customJson": "{}"
```
nikhil-zlai: nit: this shouldn't be set if empty

api/py/ai/chronon/source.py
nikhil-zlai: if kwargs is empty - set customJson to None

nikhil-zlai (changes requested):
I like the source wrappers. Can you edit docs also please?


Given that there is already a way to specify stream config alongside the topic, can we skip the api change to add customJson?

For specifying streaming params - we use topicInfo: https://github.com/airbnb/chronon/blob/bf9497bb104fa681ee12a5a77d39942bde589a27/online/src/main/scala/ai/chronon/online/DataStreamBuilder.scala#L32

```
case class TopicInfo(name: String, topicType: String, params: Map[String, String])

 topic="kafka://topic_name/schema=my_schema/host=X/port=Y" will parse into 

TopicInfo(topic_name, kafka, {schema: my_schema, host: X, port Y})
```

pyalex:
Hey @nikhil-zlai , thanks for the review!
There's more use for those extra attributes, than just Kafka host and port. For example, I want to store the Avro JSON schema near the source definition and attach it to the source. Or specify all kinds of Kafka consumer properties.

TopicInfo has limited usage since it makes `/` and `=` as special symbols, and if I were to add anything encoded with base64 to this topic string, it would simply break.

api/py/ai/chronon/source.py
pyalex: Can you please elaborate? I didn't get the idea. Didn't we specify all existing event source params in the function signature already?
```
def EventSource(
    table: str,
    query: ttypes.Query,
    topic: Optional[str] = None,
    is_cumulative: Optional[bool] = None,
```

pyalex:
Updated docs

api/py/ai/chronon/source.py
nikhil-zlai: lets say we add a new argument on api.thrift/event_source - we would need to come here and update the args. 

if we forget to do that, the arg would silently go into custom_json. 

one way to make it loud, is to check the fields on EventSource and ensure that kwargs isn't swallowing them.

hzding621: +1 to this. i think we should protect this inside a CI, as otherwise when `kwargs` check is triggered it would've been after the change is merged. 

we can use `inspect.signature` like this

```
def test_group_by():
    ttypes_params = [camel_to_snake(param) for param in inspect.signature(ttypes.GroupBy.__init__).parameters]
    binding_params = [camel_to_snake(param) for param in inspect.signature(GroupBy).parameters]

    exceptions = ["self", "meta_data", "key_columns"]
    for ttypes_param in ttypes_params:
        if ttypes_param not in binding_params and ttypes_param not in exceptions:
            raise Exception(
                f"ttypes_params {ttypes_param} not found in binding_params {binding_params}"
            )
```

pyalex: Thanks! I added [a test](https://github.com/airbnb/chronon/pull/1016/files#diff-b90366cf4ed34c7514e1cef823870073b597b68c4e3d15274b6d2c4963a21294) verifying signatures match.

api/py/ai/chronon/source.py
pyalex: Addressed

nikhil-zlai: if kwargs is empty, set it to None?

pyalex: Fixed

pyalex:
@hzding621, please take another look

api/thrift/api.thrift
```diff
@@ -68,6 +68,11 @@ struct EventSource {
     * If each new hive partition contains not just the current day's events but the entire set of events since the begininng. The key property is that the events are not mutated across partitions.
     **/
     4: optional bool isCumulative
+
+    /**
+    * Any extra attributes can be stored here. Ie, Kafka bootstrap servers for a streaming source, or AWS IAM role for accessing Iceberg table
+    **/
+    5: optional string customJson
```
pengyu-hou, line 75: can we change it to use the MetaData, because we already have `customJson` in MetaData. It is easier to wrap everything in Metadata so we don't have to add new fields in the future 
```suggestion
    5: optional MetaData metaData
```

api/thrift/api.thrift
```diff
@@ -98,6 +103,11 @@ struct EntitySource {
     * The logic used to scan both the table and the topic. Contains row level transformations and filtering expressed as Spark SQL statements.
     */
     4: optional Query query
+
+    /**
+    * Any extra attributes can be stored here. Ie, Kafka bootstrap servers for a streaming source, or AWS IAM role for accessing Iceberg table
+    **/
+    5: optional string customJson
```
pengyu-hou, line 110: Ditto, use MetaData instead of creating a new `customJson` field. 
```suggestion
    5: optional MetaData metaData
```

api/thrift/api.thrift
abbywh: @pengyu-hou I think we just went straight to `customJson` since there are a few fields in `Metadata` that don't apply to sources

```
    10: optional bool consistencyCheck
    // percentage of online serving requests to log to warehouse
    11: optional double samplePercent
```

For example

PR #1017, opened 2025-07-19
fix: allow derivations to overwrite base columns 

hzding621:
## Summary

Remove logic in validator.py that checks a derivation column name cannot be same as existing base column. 

## Why / Goal

This restriction should be relaxed such that a base column's name can be reused. A common use case of this is to support imputation of default values, for example:

```
        Derivation(
            name="group_by_name_col_name",
            expression="COALESCE(group_by_name_col_name, -1L)"
        )
```

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Reviewers

@yuli-han @airbnb/airbnb-chronon-maintainers

### DIFF
--- a/api/py/ai/chronon/repo/validator.py
+++ b/api/py/ai/chronon/repo/validator.py
@@ -317,13 +317,7 @@ class ChrononRepoValidator(object):
                         )
                     )
             if derivation.name != "*":
-                # Do not validate the name conflict for keys
-                if derivation.name in derived_columns and derivation.name not in key_cols:
-                    errors.append(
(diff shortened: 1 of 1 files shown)


yuli-han (approved):

hzding621:
actually the projection logic doesn't handle it correctly now: https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Extensions.scala#L1082

`*` + rename can be handled. overwriting a base column is not handled properly, so there is a real conflict.

PR #1027, opened 2025-08-25
[CHIP-9] Model transform batch job interface + orchestration

hzding621:
## Summary

implement model transform batch job:
- Add interface to `ModelBackend` that triggers model inference batch job 
- Orchestration logic in `ModelTransformBatchJob` which handles compute-range calculations 
- Update join bootstrap logic to handle model fields 

## Why / Goal

Support offline job for joins with model transforms 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Reviewers

@airbnb/airbnb-chronon-maintainers @pengyu-hou

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/ModelTransformBatchJob.scala
+++ b/spark/src/main/scala/ai/chronon/spark/ModelTransformBatchJob.scala
@@ -1,7 +1,23 @@
+/*
+ *    Copyright (C) 2025 The Chronon Authors.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
 package ai.chronon.spark
 import ai.chronon.api.Extensions.{JoinOps, MetadataOps, ModelTransformOps, SourceOps}
 import ai.chronon.api.ModelTransform
-import ai.chronon.online.ModelBackend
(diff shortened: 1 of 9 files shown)


spark/src/main/scala/ai/chronon/spark/SemanticHashUtils.scala
```diff
@@ -142,7 +142,11 @@ object SemanticHashUtils {
     val derivedChanges = oldSemanticHash.get(join.derivedKey) != newSemanticHash.get(join.derivedKey)
     // TODO: make this incremental, retain the main table and continue joining, dropping etc
     val mainTable = if (partsToDrop.nonEmpty || added.nonEmpty || derivedChanges) {
-      Some(join.metaData.outputTable)
+      if (join.hasModelTransforms) {
```
pengyu-hou, line 145: 👍

pengyu-hou (approved):

spark/src/main/scala/ai/chronon/spark/ModelTransformBatchJob.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private def writeFinalOutput(outputDf: DataFrame): Unit = {
+    val tableProperties = Option(join.metaData.tableProperties)
+      .map(_.toScala.toMap)
+      .getOrElse(Map.empty[String, String])
+
+    tableUtils.insertPartitions(
+      outputDf,
+      join.metaData.outputTable,
```
yuli-han, line 61: Is the model transform batch job writing to the join output table? Wander how does this work with the join backfill job, are these two jobs writing to the same table?

hzding621: @yuli-han when a join has model transforms, then its join backfill job will instead write to a `pre_mt` table: https://github.com/airbnb/chronon/blob/v0.0.106/spark/src/main/scala/ai/chronon/spark/JoinBase.scala#L49-L53

the `pre_mt` table will contain data at the stage of derivations, and this data becomes input to the stage of model transforms, whose output will be the final output.

yuli-han: Got it. Thanks for clarification

yuli-han (approved):

PR #1031, opened 2025-08-28
Add ExternalSourceFactory to Chronon

Hui-Cheng-AirBnb:
## Summary

This PR introduces ExternalSourceFactory to ExternalSource. The factory provides:
 - A factory name
 - Factory parameters

With this addition, users can apply the factory pattern to create and register handlers dynamically at runtime using factory configurations.

This enhancement does not replace the existing mechanism where handlers are explicitly registered for external sources. Instead, it complements it, allowing users to choose between explicit registration or the factory pattern depending on their implementation needs.

## Why / Goal

Currently, users can only register external source handlers explicitly, which limits flexibility and requires code changes for every new handler. By introducing ExternalSourceFactory, handlers can be created and registered dynamically at runtime based on configuration.

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers


online/src/main/scala/ai/chronon/online/Api.scala
```diff
@@ -265,4 +268,91 @@ abstract class Api(userConf: Map[String, String]) extends Serializable {
     new Consumer[LoggableResponse] {
       override def accept(t: LoggableResponse): Unit = logResponse(t)
     }
+
+  // Create a MetadataStore instance for accessing join configurations
+  private lazy val metadataStore: MetadataStore = 
+    new MetadataStore(genKvStore, Constants.ChrononMetadataKey, 10000)
```
hzding621: instead of creating new `metadataStore`, i think we should reuse Fetcher (which extends MetadataStore), as MetadataStore is resource heavy, with its thread pools and ttl caches.

hzding621 (commented):
@Hui-Cheng-AirBnb thank you for leading this change! 

For external source that needs to be dynamically registered at runtime, I think we should also update the logic inside `ExternalSourceRegistry#fetchRequests` : 
- [here](https://github.com/airbnb/chronon/blob/main/online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala#L62) we dispatch to handler based on external source name. We should handle factory-based external source differently.

online/src/test/scala/ai/chronon/online/test/ExternalSourceFactoryTest.scala
```diff
@@ -0,0 +1,322 @@
...
+    
+    // Execute and expect exception
+    val exception = intercept[IllegalArgumentException] {
+      api.registerExternalSources(Seq("test-join"))
+    }
+    
+    assertTrue("Exception should mention null factoryParams", 
+      exception.getMessage.contains("factoryParams is null"))
+  }
+}
```
hzding621, line 246: I'd recommend also creating end-to-end test that has Fetcher fetching a factory-based external source, similar to [ExternalSourcesTest](https://github.com/airbnb/chronon/blob/main/spark/src/test/scala/ai/chronon/spark/test/ExternalSourcesTest.scala)

Hui-Cheng-AirBnb:
> @Hui-Cheng-AirBnb thank you for leading this change!
> 
> For external source that needs to be dynamically registered at runtime, I think we should also update the logic inside `ExternalSourceRegistry#fetchRequests` :
> 
> * [here](https://github.com/airbnb/chronon/blob/main/online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala#L62) we dispatch to handler based on external source name. We should handle factory-based external source differently.

online/src/main/scala/ai/chronon/online/Api.scala
Hui-Cheng-AirBnb: @hzding621 Thanks for the comments! One question about "we dispatch to handler based on external source name. We should handle factory-based external source differently." Inside the new registerExternalSources function, handlers are created from the user registered factory classes and added to the ExternalSourceRegistry handlerMap. I think that'd handle the case?

hzding621: @Hui-Cheng-AirBnb the `registerExternalSources` looks like something that will be called by the user? I think we should handle the case when user doesn't call `registerExternalSources` at init time. Because it might not always be possible to pre-determine all joins and external sources that a service needs to fetch during init time; what the user can determine is all the potential ES factory names, but not the individual ES's and their corresponding ES factory params. 

For such scenarios, the request flow should look something like this:
- At service init time, register the ES factory in api needed by the service
- At request time, if it's an factory-based ES, check if the ES handler has already been registered, if not, use the factory to construct and register it.

Hui-Cheng-AirBnb: Good points! Thanks for the context

api/thrift/api.thrift
```diff
@@ -105,10 +105,16 @@ struct EntitySource {
     4: optional Query query
 }
 
+struct ExternalSourceFactoryConfig {
```
pengyu-hou, line 112: maybe add a comment for this struct

online/src/main/scala/ai/chronon/online/Api.scala
pengyu-hou: Plus one to Haozhen's comment. `Api.scala` is mostly used for KV store implementations and let's not have a Metadata Store here.

online/src/test/scala/ai/chronon/online/test/ExternalSourceFactoryTest.scala
pengyu-hou: Plus one to Haozhen's comment for end-to-end test. And maybe we can remove redundant tests and make the tests more concise here in `online/src/test/scala/ai/chronon/online/test/ExternalSourceFactoryTest.scala`.

pkundurthy (approved):

online/src/main/scala/ai/chronon/online/Api.scala
```diff
@@ -265,4 +268,35 @@ abstract class Api(userConf: Map[String, String]) extends Serializable {
...
+    * from external sources, and stores this information in the ExternalSourceRegistry.
+    *
+    * @param joins Sequence of join names to process
+    */
+  def registerExternalSources(joins: Seq[String]): Unit = {
+    val joinOps = joins.map(joinName => getJoinConfiguration(joinName))
+    joinOps
+      .map(_.join)
+      .foreach(join => externalRegistry.addHandlersByFactory(Option(join.getOnlineExternalParts).map(_.asScala.toSeq)))
+  }
```
pengyu-hou: curious if we could combine these two methods. Because `getJoinConf` method already has error logging and metric collection, we may not need it again here. So we can also remove logger in Api.scala probably.

online/src/main/scala/ai/chronon/online/Api.scala
```diff
@@ -265,4 +268,35 @@ abstract class Api(userConf: Map[String, String]) extends Serializable {
...
+
+  /**
+    * Register external sources from join configurations.
+    * This function reads configurations for each join, extracts FactoryConfig information
+    * from external sources, and stores this information in the ExternalSourceRegistry.
+    *
+    * @param joins Sequence of join names to process
+    */
+  def registerExternalSources(joins: Seq[String]): Unit = {
```
pengyu-hou: even though ES is only available in Join, but let's still add a join to the method name. 
```suggestion
  def registerJoinExternalSources(joins: Seq[String]): Unit = {
```

online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala
```diff
@@ -90,4 +147,29 @@ class ExternalSourceRegistry extends Serializable {
...
+          case Some(_) =>
+            try {
+              addHandlersByFactory(externalParts)
+              handlerMap.get(name) // Check again after registration attempt
+            } catch {
+              case _: Exception => None
+            }
+          case None => None
+        }
+    }
```
pengyu-hou: maybe we can simplify the code this way
```suggestion
    handlerMap.get(name).orElse {
      externalParts.flatMap { _ =>
        try {
          addHandlersByFactory(externalParts)
          handlerMap.get(name) // Check again after registration attempt
        } catch {
          case _: Exception => None
        }
      }
    }
```

online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala
```diff
@@ -41,39 +40,97 @@ class ExternalSourceRegistry extends Serializable {
...
+    * Add external source handlers using factory from external parts.
+    * This function processes external parts, extracts FactoryConfig information
+    * from external sources, and creates handlers using registered factories.
+    *
+    * @param externalParts Optional sequence of ExternalPart objects to process
+    */
+  def addHandlersByFactory(externalParts: Option[Seq[ExternalPart]]): Unit = {
+    // Check if external parts are provided
+    externalParts.foreach { externalParts =>
+      externalParts.foreach { externalPart =>
```
pengyu-hou: looks like we are calling `externalParts` in two nested for loops. Is it intended?

online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala
```diff
@@ -41,39 +40,97 @@ class ExternalSourceRegistry extends Serializable {
     result
   }
 
+  val handlerFactoryMap: mutable.Map[String, ExternalSourceFactory] = {
+    val result = new mutable.HashMap[String, ExternalSourceFactory]()
```
pengyu-hou: Probably we should use `ConcurrentHashMap` for both ` handlerMap` and `handlerFactoryMap` given we are increasing the usage of it.

online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala
Hui-Cheng-AirBnb: The first one is to get the value from Option[]. Agree the names and the loops can be confusing. I updated the code to make it clearer. Thanks for the comment!

online/src/main/scala/ai/chronon/online/ExternalSourceRegistry.scala
Hui-Cheng-AirBnb: Good point!

PR #1032, opened 2025-08-28
Fix over counting due to bathIR mutation when enabling cache

mickjermsurawong-openai:
## Summary

- Upon enabling cache, we re-use final bath IR.
- Currently, an edge case when collapsed value is null, we mutate batchIR and result in over-counting
- This PR clone the hop from IR tail so we don't mutate it.

## Test Plan

- Verified that before the fix tests failed, IR is mutated and we actually return same instance of map
- Added unit test.

## Checklist
- [ ] Documentation update

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/SawtoothMutationAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/SawtoothMutationAggregator.scala
@@ -160,7 +160,7 @@ class SawtoothMutationAggregator(aggregations: Seq[Aggregation],
           if ((batchEndTs - window.millis) + tailBufferMillis > hopStart && hopStart >= queryTail) {
             val hopVal = hopIr(baseIrIndices(i))
             // When the first relevant hop is null (the collapsed is null),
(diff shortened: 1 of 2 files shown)


nikhil-zlai (approved):
very neat! thanks mick!

pengyu-hou (approved):
Thanks @mickjermsurawong-openai !

aggregator/src/main/scala/ai/chronon/aggregator/windowing/SawtoothMutationAggregator.scala
```diff
@@ -158,7 +158,14 @@ class SawtoothMutationAggregator(aggregations: Seq[Aggregation],
           val hopIr = hopIrs(idx)
           val hopStart = hopIr.last.asInstanceOf[Long]
           if ((batchEndTs - window.millis) + tailBufferMillis > hopStart && hopStart >= queryTail) {
-            relevantHops += hopIr(baseIrIndices(i))
+            val hopVal = hopIr(baseIrIndices(i))
+            // When first relevant hop is null, there the collapsed is null,
```
pengyu-hou:
```suggestion
            // When first relevant hop is null, the collapsed is null,
```

PR #1041, opened 2025-09-25
Fix logging avro serde for struct columns 

hzding621:
## Summary

When multiple struct columns exist in the same avro schema (e.g. same join) that contain sub-columns with the same name, a suffix REPEATED_{i} is added. This PR fixes an issue in this scenario where the suffix is not handled, causing logging and log-flattener job failures. 

## Why / Goal

Fix logging and log-flattener job failures. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Reviewers

### DIFF
--- a/spark/src/test/scala/ai/chronon/spark/test/SchemaEvolutionTest.scala
+++ b/spark/src/test/scala/ai/chronon/spark/test/SchemaEvolutionTest.scala
@@ -176,15 +176,24 @@ class SchemaEvolutionTest extends TestCase {
       "struct_schema",
       Array(
         StructField("id", LongType),
-        StructField(name, ListType(StructType(name = "s", fields = Array(StructField("i", LongType))))),
+        StructField(
+          name,
+          StructType(
+            name = "struct_schema_2",
+            fields =
+              Array(StructField("id", LongType),
+                    StructField("context",
+                                StructType(name = "struct_schema_3", fields = Array(StructField("id", LongType)))))
+          )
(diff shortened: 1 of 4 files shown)


spark/src/test/scala/ai/chronon/spark/test/SchemaEvolutionTest.scala
```diff
@@ -171,6 +171,48 @@ class SchemaEvolutionTest extends TestCase {
...
   }
 
+  def createStructGroupBy(namespace: String, spark: SparkSession, name: String): GroupByTestSuite = {
+    val schema = StructType(
+      "struct_schema",
+      Array(
+        StructField("id", LongType),
+        StructField(name, ListType(StructType(name = "s", fields = Array(StructField("i", LongType))))),
+        StructField("ds", StringType)
+      )
```
pengyu-hou, line 190: Can we add another nested struct test case with potential duplicate names like below? 
```
{ event_type,
 user_id, 
context: {user_id, timestamp}
}
```

pengyu-hou (approved):
Thanks for the fix. Left one comment for unit test.

spark/src/test/scala/ai/chronon/spark/test/SchemaEvolutionTest.scala
hzding621: Great idea

PR #1044, opened 2025-09-29
Add offlineGroupBy option for external source

Hui-Cheng-AirBnb:
## Summary

Add and validate offline groupby option for external source. This is the first part of the [task](https://app.asana.com/1/150975571430/project/1202600360134537/task/1211028295382896?focus=true) to support backfill for external sources. This PR does:
1. Add offline GroupBy to ExternalSource 
2. Validate the GroupBy schema is compatible with the External source.
3. Add offline GroupBy as part of the join parts for offline backfill

## Why / Goal

We want to introduce a first-level abstraction for External source to support offline backfill. The idea is that an ExternalSource now can take in an optional GroupBy definition. During offline backfill, Chronon will dispatch to the GroupBy for backfill. 

## Test Plan

- [x] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

========= Test configs =======
Test configs are at  https://git.musta.ch/airbnb/ml_models/pull/27812 

========= Export schema with External offline groupby =========
```
python \
    run_on_bigair.py run \
    production/joins/zipline_test/test_online_join_external.v2 \
    --envs "VERSION:hui-cheng-ext-gb-1-0.0.110-SNAPSHOT" \
    -- \
    --ds=2025-10-12 \
    --mode=analyze \
    --skip-timestamp-check \
    --skip-table-permission-check \
    --export-schema
```
* BigAir UI link: https://mli.airbnb.tools/bigair/workflow-run/4DPr4cRADSwqzr6dW3cXD8 
* airflow_url: https://airflow-neon.a.musta.ch/tree?dag_id=chronon_runpy_wf.hui_cheng.4DPr4cRADSwqzr6dW3cXD8 
Screenshot 2025-10-17 at 9 16 47 AM

========= Backfill join with External offline groupby ===========
```
python \
    run_on_bigair.py run \
    production/joins/zipline_test/test_online_join_external.v2 \
    --envs "VERSION:hui-cheng-ext-gb-1-0.0.110-SNAPSHOT" \
    -- \
    --ds=2025-10-12 \
    --mode=backfill
```
* BigAir UI link: https://mli.airbnb.tools/bigair/workflow-run/30pM325mrOoLRgCNEU9Xva 
* airflow_url: https://airflow-neon.a.musta.ch/tree?dag_id=chronon_runpy_wf.hui_cheng.30pM325mrOoLRgCNEU9Xva 

https://superset.a.musta.ch/sqllab/p/Oa8yamlbN8K/

======== Backfill join WITHOUT External offline groupby ========
```
python \
    run_on_bigair.py run \
    production/joins/zipline_test/test_online_join_small.v2 \
    --envs "VERSION:hui-cheng-ext-gb-1-0.0.110-SNAPSHOT" \
    -- \
    --ds=2025-10-12 \
    --mode=backfill
```
* BigAir UI link: https://mli.airbnb.tools/bigair/workflow-run/5OHqpaOMUsyqMZzpJDSSRV 
* airflow_url: https://airflow-neon.a.musta.ch/tree?dag_id=chronon_runpy_wf.hui_cheng.5OHqpaOMUsyqMZzpJDSSRV 

https://superset.a.musta.ch/sqllab/p/gKx6rAQ2kxz/ 

Compare with backfilled table with official Chronon build
https://superset.a.musta.ch/sqllab/p/4RJ9e9zBwJA/ 

========= Export schema WITHOUT External offline groupby =====
```
python \
    run_on_bigair.py run \
    production/joins/zipline_test/test_online_join_small.v2 \
    --envs "VERSION:hui-cheng-ext-gb-1-0.0.110-SNAPSHOT" \
    -- \
    --ds=2025-10-12 \
    --mode=analyze \
    --skip-timestamp-check \
    --skip-table-permission-check \
    --export-schema
```
* BigAir UI link: https://mli.airbnb.tools/bigair/workflow-run/38Q3SKmezh9Xt4dgZPJMju
* airflow_url: https://airflow-neon.a.musta.ch/tree?dag_id=chronon_runpy_wf.hui_cheng.38Q3SKmezh9Xt4dgZPJMju 

Check schema matches with the schema exported from official package
https://superset.a.musta.ch/sqllab/p/6B3NRO5A13R/

## Checklist
- [ ] Documentation update

### DIFF
--- a/api/py/ai/chronon/join.py
+++ b/api/py/ai/chronon/join.py
@@ -58,7 +58,7 @@ def JoinPart(
         components like GroupBys.
     """
     # Automatically set the GroupBy name if not already set
-    _auto_set_group_by_name(group_by, context="JoinPart")
+    _auto_set_group_by_name(group_by)
 
     if key_mapping:
         utils.check_contains(key_mapping.values(), group_by.keyColumns, "key", group_by.metaData.name)
@@ -171,7 +171,7 @@ def ExternalSource(
 
     # Automatically set the name for offline_group_by if not already set
     if offline_group_by is not None:
-        _auto_set_group_by_name(offline_group_by, context="ExternalSource")
+        _auto_set_group_by_name(offline_group_by)
 
     factory_config = None
     if factory_name is not None or factory_params is not None:
@@ -646,18 +646,13 @@ def Join(
     )
 
 
-def _auto_set_group_by_name(group_by: api.GroupBy, context: str = "GroupBy") -> None:
+def _auto_set_group_by_name(group_by: api.GroupBy) -> None:
     """
     Automatically set the GroupBy name by finding its source module using garbage collection.
     This is used by both JoinPart and ExternalSource to automatically name GroupBys.
 
     :param group_by: The GroupBy object to set the name for
-    :param context: Context string for error messages (e.g., "JoinPart", "ExternalSource")
     """
-    if group_by.metaData.name:
-        # Name already set, nothing to do
-        return
-
     # Save and restore __import__ to preserve original behavior
     import_copy = __builtins__["__import__"]
 
@@ -671,15 +666,15 @@ def _auto_set_group_by_name(group_by: api.GroupBy, context: str = "GroupBy") -> 
                 break
 
         if group_by_module_name:
-            logging.debug(f"{context}: group_by's module info from garbage collector {group_by_module_name}")
+            logging.debug("group_by's module info from garbage collector {}".format(group_by_module_name))
             group_by_module = importlib.import_module(group_by_module_name)
             __builtins__["__import__"] = eo.import_module_set_name(group_by_module, api.GroupBy)
         else:
             if not group_by.metaData.name:
-                logging.error(f"{context}: No group_by file or custom group_by name found")
+                logging.error("No group_by file or custom group_by name found")
                 raise ValueError(
-                    f"[{context}] Must specify a group_by name if group_by is not defined in separate file. "
-                    "You can set it via GroupBy(metaData=MetaData(name='team.file.variable_name'))"
+                    "[GroupBy] Must specify a group_by name if group_by is not defined in separate file. "
+                    "You may pass it in via GroupBy.name. \n"
                 )
     finally:
         # Reset before next run


api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -745,9 +745,18 @@ object Extensions {
   }
 
   implicit class JoinPartOps(joinPart: JoinPart) extends JoinPart(joinPart) {
-    lazy val fullPrefix = (Option(prefix) ++ Some(groupBy.getMetaData.cleanName)).mkString("_")
+    lazy val fullPrefix = {
+      Option(groupBy.getMetaData.customJsonLookUp("customizedFullPrefix"))
+        .map(_.toString)
+        .getOrElse((Option(prefix) ++ Some(groupBy.getMetaData.cleanName)).mkString("_"))
```
hzding621: I assume `customizedFullPrefix` is used in order to map the GroupBy feature names to match with ExternalSource feature names? 

As an alternative solution, I think rather than including `customizedFullPrefix` in the GroupBy definition, maybe we can allow users to specify a mapping inside the ExternalSource definition (sorry I didn't previously think through this). 

```
struct ExternalSource {
    1: optional MetaData metadata
    2: optional TDataType keySchema
    3: optional TDataType valueSchema
    4: optional ExternalSourceFactoryConfig factoryConfig
    5: optional ExternalSourceOfflineConfig offlineConfig
}

struct ExternalSourceOfflineConfig {
    1: optional GroupBy offlineGroupBy
    2: optional map<string, string> nameMappings
}
```

I think this is better because we don't need to include ExternalSource-specific information into the GroupBy object, which can be standalone.

hzding621: Having `ExternalSourceOfflineConfig` also allows us introduce new params related to offline in a more organized way.

hzding621: Though downside to this approach is that you cannot simply reuse `JoinPart` since `JoinPart` doesn't capture value mappings easily

hzding621: Another solution is to rely on `GroupBy`'s own `derivations` layer, which can arbitrarily rename the output fields of `GroupBy`, so that we also don't have to rely on a customJson field.

spark/src/test/scala/ai/chronon/spark/test/ExternalSourceBackfillTest.scala
```diff
@@ -0,0 +1,420 @@
...
+    // Create Join and compute
+    val endPartition = monthAgo
+    val join = new Join(joinConf = joinConf, endPartition = endPartition, tableUtils)
+    val computed = join.computeJoin(Some(10))
+
+    // Verify results
+    assertNotNull("Computed result should not be null", computed)
+    assertTrue("Result should have rows", computed.count() > 0)
+
+    // Verify that external source columns are present
```
hzding621: Could you also verify that ES columns have non-null data? Current behavior is that Join job will always create ES columns but leave them as null (unless bootstrapped).

api/src/main/scala/ai/chronon/api/Extensions.scala
Hui-Cheng-AirBnb: @hzding621 , thanks for the comments! I removed the custom JSON and introduced ExternalJoinPart to address the prefix issue. I also explored naming-convention-based column mapping that we discussed, but since fullPrefix is used across many components, that would cause wide-ranging changes. Centralizing prefix handling in one place seems cleaner and safer. PTAL and let me know if you see any concerns.

api/src/main/scala/ai/chronon/api/ExternalJoinPart.scala
```diff
@@ -0,0 +1,9 @@
+package ai.chronon.api
```
hzding621, line 17: nit: add `Copyright (C)` docs at top for new files

api/py/ai/chronon/repo/compile.py
```diff
@@ -148,6 +149,10 @@ def extract_and_convert(chronon_root, input_path, output_root, debug, force_over
                     else:
                         offline_gbs.append(jp.groupBy.metaData.name)
 
+                # Extract and always materialize online GroupBys from external parts
+                external_offline_gbs = _extract_external_part_offline_group_bys(obj, team_name, teams_path)
+                online_group_bys.update(external_offline_gbs)
```
hzding621: why do we need to materialize offline group_bys from external sources?

Hui-Cheng-AirBnb: The scenario is when users update the groupby used for offline backfill and compile join config, we'd like to also compile the updated groupby - similar to what happens today for regular groupby. So it is consistent behavior for groupbys used for regular join part of external source.

hzding621: I see. I think the naming of the method is causing the confusion. 

A group_by has a boolean flag called online. You can directly compile a group_by, which generates a json file in production/group_bys/... ; You can also compile a join, which generates a json file in production/joins/..., and in addition, for each group_by inside the join, if the group_by.online=True, it also compiles the group_by, and generates the production/group_bys/... . This `online_group_bys` is for this purpose. 

For more context, we don't need to materialize offline group_bys. The reason is that to enable online group_bys, we set up separate pipelines (such as batch upload, streaming jobs). These happen at group_by level, so we materialize the production/group_bys/.. files for this purpose. For offline group_bys, we don't need to create separate pipelines for each group_by. Rather, the join handles the backfill. Each join materialized file has the full copy of the group_bys logic.

Hui-Cheng-AirBnb: Thanks for the context! It makes sense not to materialize the external backfill groupby. I'll update to remove this part.

api/py/ai/chronon/join.py
```diff
@@ -180,10 +162,17 @@ def ExternalSource(
         creating the external source handler.
     :param factory_params: Optional parameters to pass to the factory when
         creating the handler.
+    :param offline_group_by: Optional GroupBy configuration to be used for
+        offline backfill computation. When provided, enables point-in-time
```
pengyu-hou:
```suggestion
        offline backfill computation consuming from a mutation table for the service data. When provided, enables point-in-time
```

spark/src/test/scala/ai/chronon/spark/test/ExternalSourceBackfillTest.scala
pengyu-hou: Can you write a sql query to sum the external offline group by and compare it with the backfilled result from the join? 

Here is one example: https://github.com/airbnb/chronon/blob/main/spark/src/test/scala/ai/chronon/spark/test/JoinTest.scala#L330-L344

spark/src/test/scala/ai/chronon/spark/test/ExternalSourceBackfillTest.scala
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    // Verify that regular JoinPart columns have non-null data
+    val joinPartColumns = computed.columns.filter(_.startsWith("session_"))
+    assertTrue("Should have at least one JoinPart column", joinPartColumns.nonEmpty)
+    joinPartColumns.foreach { col =>
+      val nonNullCount = computed.filter(s"$col IS NOT NULL").count()
+      assertTrue(s"JoinPart column $col should have non-null values (found $nonNullCount non-null rows)",
+                 nonNullCount > 0)
+    }
```
pengyu-hou: similar to the previous comment. Let's create an expected data frame based on the sql query from the job, and then compare the two data frames. It will be more effective than simply check column existence and nullability.

spark/src/test/scala/ai/chronon/spark/test/ExternalSourceBackfillTest.scala
```diff
@@ -0,0 +1,437 @@
...
+    // Verify results
+    assertNotNull("Computed result should not be null", computed)
+    assertTrue("Result should have rows", computed.count() > 0)
+
+    // Verify column structure
+    val columns = computed.columns.toSet
+    assertTrue("Should contain external_user_id from left", columns.contains("external_user_id"))
+    assertTrue("Should contain request_type from left", columns.contains("request_type"))
+    assertTrue("Should contain mapped external columns",
+               columns.exists(_.startsWith("ext_mapped_")))
```
pengyu-hou: ditto

Hui-Cheng-AirBnb: Thanks @pengyu-hou . I've updated the tests to comparing the dataframes. PTAL thanks!

pengyu-hou: Hi, thanks for improving the unit test! Can you generate the test data using [DataframeGen](https://github.com/airbnb/chronon/blob/main/spark/src/test/scala/ai/chronon/spark/test/DataFrameGen.scala)? Here is one [example 1](https://github.com/airbnb/chronon/blob/main/spark/src/test/scala/ai/chronon/spark/test/JoinBasicTest.scala#L66-L89)?

There is a comparison utility already so you don't have to sort the data frame by yourself: https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/Comparison.scala

Then we could write an expected dataframe by querying the generated input data like this [example](https://github.com/airbnb/chronon/blob/main/spark/src/test/scala/ai/chronon/spark/test/JoinBasicTest.scala#L215-L242)

Sorry for being a bit pedantic on this. We prefer randomly generated data to mimic the real world data over the deterministic data. 

If this is blocking your change, we could take it as a follow up item.

Hui-Cheng-AirBnb: Thanks Pengyu for the suggestion! I've updated the PR following the example you provided.

pengyu-hou: Thanks so much for your patience! Stamped!

PR #1060, opened 2025-10-24
CI Speed, Test Flakes

abbywh:
## Summary

* Split tests
* Isolate Fetcher tests
* Port a known bugfix (schema should come from limit 0 not limit 1)

## Why / Goal

CI needs to go fast and be correct

Fetcher/spark tests flakes are significantly improved

Builds in 10 minutes instead of 30

## Test Plan

- [ ] Added Unit Tests
- [ ] Integration tested

## Checklist
- [ ] Documentation update


abbywh:
should split join test and maybe a few others but that can be done later

spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -448,7 +449,8 @@ abstract class JoinBase(joinConf: api.Join,
     } else {
       // Register UDFs. `setups` from entire joinConf are run due to BootstrapInfo computation
       joinConf.setups.foreach(tableUtils.sql)
-      val leftSchema = leftDf(joinConf, unfilledRanges.head, tableUtils, limit = Some(1)).map(df => df.schema)
+      val leftSchema =
+        leftDf(joinConf, unfilledRanges.head, tableUtils, allowEmpty = true, limit = Some(0)).map(df => df.schema)
```
pengyu-hou, line 453: this is actually great. I did not know we can simply fetch schema if limit the rows to 0.

PR #1061, opened 2025-10-29
update join backfill wf

hzding621:
## Summary

Add dependency between left and right if there is no join_parts in the join. 

## Why / Goal

Fix DAG topology 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

## Reviewers


yuli-han (approved):
If there is no join parts the left table is the same as final table? Wander what would be the use case for this?

pengyu-hou (approved):

pengyu-hou:
> If there is no join parts the left table is the same as final table? Wander what would be the use case for this?

I have the same question.

PR #1063, opened 2025-11-07
Update ChrononRunDS assert criteria

yogeshpv:
## Summary
- Updated assertion validation from interesectedRange to queryRange
- Fixed macro replacement to wrap dates in quotes
>    Before: ParametricMacro(Constants.ChrononRunDs, _ => queryRange.start)
>    After: ParametricMacro(Constants.ChrononRunDs, _ => s"'${queryRange.start}'")
>  This ensures the generated SQL contains '2023-01-01' (string literal) instead of 2023-01-01 (arithmetic)

## Why / Goal
The `ChrononRunDS` allows the use of run date in `groupby`. The update involves modifying the assertion logic for the `singleDay` condition on `queryRange` to ensure it functions correctly when the source query covers exactly one day's data. The window functions modify the `intersectedRange`, so the system relies on the original queryRange for this assertion.

## Test Plan
Test was available in #592 

## Reviewers
@pengyu-hou 
@nikhil-zlai

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ And then:
 spark-sql> SELECT user_id, quickstart_returns_v1_refund_amt_sum_30d, quickstart_purchases_v1_purchase_price_sum_14d, quickstart_users_v1_email_verified from default.quickstart_training_set_v1 limit 100;
 ```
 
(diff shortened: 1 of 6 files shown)


pengyu-hou:
cc @hzding621 can you take a look

PR #1065, opened 2025-11-11
Add a vector average operation for computing an element-wise average

rgruener:
## Summary
This adds a new aggregation for computing a running vector average of a field. It utilizes a running average that is generally safe from overflow.

## Why / Goal
This is very useful for computing an average embedding vector.

## Test Plan
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

This is also currently in use within the Netflix internal fork.

## Checklist
- [ ] Documentation update

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Analyzer.scala
@@ -236,7 +236,8 @@ class Analyzer(tableUtils: TableUtils,
       .mkString("\n")
   }
 
-  // TODO: Add check for element wise operations (e.g. dimensions)
+  // TODO: Add check for element wise operations (e.g. dimensions and null-ness)
+  // More details at: https://github.com/airbnb/chronon/pull/1065#discussion_r2524393195
   def analyzeGroupBy(groupByConf: api.GroupBy,
                      prefix: String = "",
                      includeOutputTableName: Boolean = false,


nikhil-zlai (changes requested):
refer to the message prior

api/thrift/api.thrift
```diff
@@ -222,9 +222,16 @@ struct Aggregation {
...
+    This is an additional layer of aggregation. You can key a group_by by user, and bucket a "item_view" count by "item_category". This will produce one row per user, with column containing map of "item_category" to "view_count". You can specify multiple such buckets at once
     */
     5: optional list<string> buckets
+
+    /**
+    When set to true and inputColumn is an array/list type, applies the operation element-wise across the arrays.
+    For example, AVERAGE with tensorElementWiseOperation=true on [[1,2,3], [4,5,6]] produces [2.5, 3.5, 4.5].
+    This allows any operation (SUM, AVERAGE, MAX, MIN, etc.) to work on tensors.
+    */
+    6: optional bool tensorElementWiseOperation
```
nikhil-zlai: can we shorten the name? to "tensor" or "element_wise" perhaps?

rgruener: sure, was just trying to make it a bit more explicit

aggregator/src/main/scala/ai/chronon/aggregator/row/TensorColumnAggregator.scala
```diff
@@ -0,0 +1,108 @@
...
+
+package ai.chronon.aggregator.row
+
+import ai.chronon.aggregator.base.SimpleAggregator
+import ai.chronon.api.Row
+
+import java.util
+import scala.util.ScalaJavaConversions.IteratorOps
+
+class TensorColumnAggregator[Input, IR, Output](agg: SimpleAggregator[Input, IR, Output],
```
nikhil-zlai: we could rename here also for consistency - ElementWiseAggregator

nikhil-zlai (approved):
awesome work! approving.

i added one more comment - it will be good to do a pass over the changes to remove all usages of "tensor" with "elementWise"

aggregator/src/main/scala/ai/chronon/aggregator/row/ElementWiseAggregator.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def tensorIterator(inputRow: Row): Iterator[(Int, Input)] = {
+    val inputVal = inputRow.get(columnIndices.input)
+    if (inputVal == null) return null
+    inputVal match {
+      case inputJList: util.ArrayList[Any] =>
+        inputJList
+          .iterator()
+          .toScala
+          .zipWithIndex
```
nikhil-zlai: this is going to be expensive I think. We have noticed tuple creation taking up room in the flamegraphs in the past.

might be worth converting this to a while loop and taking a consumer func that consumes (idx, typeValue).

aggregator/src/main/scala/ai/chronon/aggregator/row/ElementWiseAggregator.scala
```diff
@@ -0,0 +1,118 @@
...
+        i += 1
+      }
+      typedRow.update(columnIndices.output, ir)
+      ir
+    } else {
+      val existing = irVal.asInstanceOf[util.ArrayList[Any]]
+      // Verify dimensions match
+      if (existing.size() != tensorSize) {
+        throw new IllegalStateException(
+          s"Tensor dimensions must match: existing=${existing.size()}, input=$tensorSize"
```
pengyu-hou, line 93: We have a job called Analyzer which will perform some quick checks before starting the heavy computations. Because element wise has a strong assertion on the size, can you help add a validation check logic similar to [runTimestampChecks](https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/Analyzer.scala#L686)? We can sample a few rows and make sure the tensor size is all the same.

pengyu-hou (commented):
Thanks a lot for the change! One minor comment otherwise it looks good. 

Thanks @nikhil-zlai for the initial reviews!

aggregator/src/main/scala/ai/chronon/aggregator/row/ElementWiseAggregator.scala
nikhil-zlai: this checks null-ness of the element, which i am afraid will be very expensive for analyzer. if the dimension of the vec is too large the analyzer will take up significant room in the job

pengyu-hou: Had a discussion offline with @nikhil-zlai that the logic to sample -> check null-ness and array size can be in a follow up PR. 

@rgruener , could you add a TODO to [analyzeGroupBy](https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/Analyzer.scala#L239) and put up a follow up PR when you get a chance? Thanks.

rgruener: Sure, added the TODO

spark/src/main/scala/ai/chronon/spark/Analyzer.scala
```diff
@@ -236,6 +236,7 @@ class Analyzer(tableUtils: TableUtils,
       .mkString("\n")
   }
 
+  // TODO: Add check for element wise operations (e.g. dimensions)
```
pengyu-hou:
```suggestion
  // TODO: Add check for element wise operations (e.g. dimensions and null-ness)
  // More details at: https://github.com/airbnb/chronon/pull/1065#discussion_r2524393195
```

rgruener:
Hmm I am not familiar with the semantic-hash logic. Perhaps @pengyu-hou or @nikhil-zlai can comment on if that is a concern

nikhil-zlai:
> @rgruener Thanks for making the change! I’m wondering whether elementWise will be added to existing GroupBys/Joins when we re-compile them. It looks like it’s being added to many unit-test configs. I’m a bit concerned that if it gets added to existing GroupBys during re-compilation, it could trigger semantic-hash changes and unexpected table archiving.


it will be unset / absent by default. so sem hash should not change. 

can you point us to the unit-test configs?

api/py/ai/chronon/group_by.py
```diff
@@ -129,6 +129,7 @@ def Aggregation(
     windows: Optional[List[ttypes.Window]] = None,
     buckets: Optional[List[str]] = None,
     tags: Optional[Dict[str, str]] = None,
+    element_wise: bool = False,
```
nikhil-zlai, line 132: ooh this should be null!!

rgruener: I can fix in a follow up

rgruener: https://github.com/airbnb/chronon/pull/1068

pengyu-hou: Stamped Robert's PR.

PR #1066, opened 2025-11-12
Switch average implementation to utilize running average to prevent overflow

rgruener:
## Summary
Currently the average implementation uses a sum / count. For large aggregations this can cause an overflow of sum. This changes the implementation to utilize a running average that will not overflow.

## Why / Goal
We utilize large (global) aggregations when doing tensor computations. This guarantees average will work even with these larger aggregations

## Test Plan
- [] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
@@ -123,24 +123,67 @@ class Average extends SimpleAggregator[Double, Array[Any], Double] {
   override def irType: DataType =
     StructType(
       "AvgIr",
+      Array(StructField("sum", DoubleType), StructField("count", IntType))
+    )
+
+  override def prepare(input: Double): Array[Any] = Array(input, 1)
+
+  // mutating
+  override def update(ir: Array[Any], input: Double): Array[Any] = {
+    ir.update(0, ir(0).asInstanceOf[Double] + input)
+    ir.update(1, ir(1).asInstanceOf[Int] + 1)
+    ir
+  }
+
+  // mutating
+  override def merge(ir1: Array[Any], ir2: Array[Any]): Array[Any] = {
+    ir1.update(0, ir1(0).asInstanceOf[Double] + ir2(0).asInstanceOf[Double])
+    ir1.update(1, ir1(1).asInstanceOf[Int] + ir2(1).asInstanceOf[Int])
+    ir1
+  }
+
+  override def finalize(ir: Array[Any]): Double =
+    ir(0).asInstanceOf[Double] / ir(1).asInstanceOf[Int].toDouble
+
+  override def delete(ir: Array[Any], input: Double): Array[Any] = {
+    ir.update(0, ir(0).asInstanceOf[Double] - input)
+    ir.update(1, ir(1).asInstanceOf[Int] - 1)
+    ir
+  }
+
+  override def clone(ir: Array[Any]): Array[Any] = {
+    val arr = new Array[Any](ir.length)
+    ir.copyToArray(arr)
+    arr
+  }
+
+  override def isDeletable: Boolean = true
+}
+
+class RunningAverage extends SimpleAggregator[Double, Array[Any], Double] {
+  override def outputType: DataType = DoubleType
+
+  override def irType: DataType =
+    StructType(
+      "RunningAvgIr",
       Array(StructField("running_average", DoubleType), StructField("weight", DoubleType))
     )
 
   override def prepare(input: Double): Array[Any] = Array(input, 1.0)
 
   /**
-   * When combining averages, if the counts sizes are too close we should use a different algorithm. This
-   * constant defines how close the ratio of the smaller to the total count can be:
-   */
+    * When combining averages, if the counts sizes are too close we should use a different algorithm. This
+    * constant defines how close the ratio of the smaller to the total count can be:
+    */
   private[this] val STABILITY_CONSTANT = 0.1
 
   /**
-   * Given two streams of doubles (left, leftWeight) and (right, rightWeight) of form (mean, weighted count), calculates
-   * the mean of the combined stream.
-   *
-   * Uses a more stable online algorithm which should be suitable for large numbers of records similar to:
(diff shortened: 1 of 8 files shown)


aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -123,32 +123,40 @@ class Average extends SimpleAggregator[Double, Array[Any], Double] {
   override def irType: DataType =
     StructType(
       "AvgIr",
-      Array(StructField("sum", DoubleType), StructField("count", IntType))
+      Array(StructField("running_average", DoubleType), StructField("count", IntType))
     )
 
   override def prepare(input: Double): Array[Any] = Array(input, 1)
 
+  private def computeRunningAverage(left: Double, leftCount: Int, right: Double, rightCount: Int): Array[Any] = {
```
nikhil-zlai: there is apparently a numerical stability issue in the approach (adding two doubles directly compounds the errors in the double representation)

we need to instead do something similar to: 

https://github.com/twitter/algebird/blob/464917d6a3f2b1c9f2048e3dd38df5dd97edc965/algebird-core/src/main/scala/com/twitter/algebird/MomentsGroup.scala#L291

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -123,32 +123,40 @@ class Average extends SimpleAggregator[Double, Array[Any], Double] {
   override def irType: DataType =
     StructType(
       "AvgIr",
-      Array(StructField("sum", DoubleType), StructField("count", IntType))
+      Array(StructField("running_average", DoubleType), StructField("count", IntType))
```
nikhil-zlai: unfortunately, this might break pipelines that are already in prod. 

best way to deal with this is to add a new aggregation and leave this as is.

rgruener: The concern is that changing the implementation would introduce skew?

As a minimal fix, changing `count` to a Long would be helpful

rgruener: Though I do think adding another implementation is warranted to unblock certain use cases

nikhil-zlai: there is avro encoded data that is sitting in kvStore in the old format. the new aggregation logic will probably fail to parse it.

pengyu-hou: echo Nikhil's comment.

rgruener: I see, after a bit of further investigation I believe I will add this under `RunningAverage` since I believe we will hit overflow issues (especially with count being an INT)

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -160,6 +160,82 @@ class Average extends SimpleAggregator[Double, Array[Any], Double] {
   override def isDeletable: Boolean = true
 }
 
+class RunningAverage extends SimpleAggregator[Double, Array[Any], Double] {
```
rgruener, line 163: Instead of a new operator would it make sense to add an argument `prevent_overflow` or `running_average` to the Average operator that defaults to false?

nikhil-zlai: yeah i like that! defaults to None that gets interpreted as false - to keep the semantic hashes as they were

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -160,6 +160,82 @@ class Average extends SimpleAggregator[Double, Array[Any], Double] {
...
+  private[this] val STABILITY_CONSTANT = 0.1
+
+  /**
+    * Given two streams of doubles (left, leftWeight) and (right, rightWeight) of form (mean, weighted count), calculates
+    * the mean of the combined stream.
+    *
+    * Uses a more stable online algorithm which should be suitable for large numbers of records similar to:
+    * http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+    */
+  private def computeRunningAverage(ir: Array[Any], right: Double, rightWeight: Double): Array[Any] = {
```
nikhil-zlai, line 187: there is apparently already a getCombinedMean below in the moments stuff

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
```diff
@@ -160,6 +160,82 @@ class Average extends SimpleAggregator[Double, Array[Any], Double] {
...
+      val newAverage = newCount match {
+        case 0.0                                => 0.0
+        case newCount if newCount == leftWeight => left
+        case newCount =>
+          val scaling = rightWeight / newCount
+          if (scaling < STABILITY_CONSTANT) {
+            left + (right - left) * scaling
+          } else {
+            (leftWeight * left + rightWeight * right) / newCount
+          }
```
nikhil-zlai, line 203: I tried to do this in-place (replace logic in average directly). and it actually makes the tests fail - the average operation stops being commutative due to slight errors in the double multiply and double division. I also tried the `(lw*la + rw*ra) / (lw + rw)` - without luck.

we ended up merging the following change instead: https://github.com/zipline-ai/chronon/pull/1292

nikhil-zlai (commented):
i found some weird behavior

nikhil-zlai (approved):
approving to negate my request changes. given the loss of commutativity with the running average computation (due to double multiple / divide errors) - i don't know what is the right thing to do :-/ 

we ended up just changing the denominator to long in our fork.

aggregator/src/main/scala/ai/chronon/aggregator/base/SimpleAggregators.scala
rgruener: Good to know

PR #1067, opened 2025-11-13
Use local join conf for fetcher cli test 

pengyu-hou:
## Summary

When joinConf (Option[Join]) is passed to fetcher.fetchJoin(), the fetcher was ignoring it and always fetching the codec from the KV store. This caused issues when developers wanted to use a config (e.g., for testing modified join configurations without uploading to KV store).

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [x] gateway machines 
```
python3 ~/.local/bin/run.py --mode=fetch --conf-path <conf_path> -k '{"key":"keys"}' -t join
```
## Checklist
- [ ] Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ And then:
 spark-sql> SELECT user_id, quickstart_returns_v1_refund_amt_sum_30d, quickstart_purchases_v1_purchase_price_sum_14d, quickstart_users_v1_email_verified from default.quickstart_training_set_v1 limit 100;
 ```
 
-Note that this only selects a few columns. You can also run a `select * from default.quickstart_training_set_v1 limit 100` to see all columns, however, note that the table is quite wide and the results might not be very readable on your screen.
+Note that this only selects a few columns. You can also run a `select * from default.quickstart_training_set_v1 limit 100;` to see all columns, however, note that the table is quite wide and the results might not be very readable on your screen.
 
 To exit the sql shell you can run:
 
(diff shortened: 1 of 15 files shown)


online/src/main/scala/ai/chronon/online/Fetcher.scala
hzding621: Should we pass in joinConf for fetchExternal as well?

pengyu-hou: good catch.

online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -405,7 +410,7 @@ class Fetcher(val kvStore: KVStore,
   private def doFetchJoin(requests: scala.collection.Seq[Request],
                           joinConf: Option[api.Join] = None): Future[scala.collection.Seq[Response]] = {
     val baseValuesF = fetchBaseJoin(requests, joinConf)
-    val derivedValuesF = fetchDerivations(baseValuesF)
+    val derivedValuesF = fetchDerivations(baseValuesF, joinConf)
     val modelTransformsF = fetchModelTransforms(derivedValuesF)
```
yuli-han, line 414: do we need to add joinConf in modelTransforms?

pengyu-hou: fetchModelTransforms doesn't need changes (it already uses the codec from ResponseWithContext)

online/src/main/scala/ai/chronon/online/Fetcher.scala
pengyu-hou: Updated. @hzding621

yuli-han (approved):

PR #1070, opened 2025-11-14
[Online]Enable ds derivation in online for rename-only derivations

yuli-han:
## Summary

Currently we are enabling derivation on ds for offline(It is used for the case where the wildcard derivation is not used and ds is not included in final schema), but online derivation on ds is disabled  for rename-only derivations, which caused online-offline consistency. 

The change is to enable the ds derivation in online for rename-only derivations

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @hzding621 @Shiyinghaha


pengyu-hou:
> Can we add a test case for this?

Plus 1. We should add a new testing case.

PR #1072, opened 2025-11-17
Support global aggregations

rgruener:
## Summary
This adds support for `keys=None` to signify a global aggregation. Global aggregations can be added to a join and are then joined with each record on the left since they have no keys.

## Why / Goal
Global aggregations are extremely useful for computing sane defaults to be used later in a join coalesce.

## Test Plan
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/KvRdd.scala
+++ b/spark/src/main/scala/ai/chronon/spark/KvRdd.scala
@@ -51,9 +51,18 @@ sealed trait BaseKvRdd {
   val baseFlatSchema: StructType = StructType(keySchema ++ valueSchema)
   def flatSchema: StructType = if (withTime) StructType(baseFlatSchema :+ timeField) else baseFlatSchema
   def flatZSchema: api.StructType = flatSchema.toChrononSchema("Flat")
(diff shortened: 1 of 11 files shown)


spark/src/main/scala/ai/chronon/spark/KvRdd.scala
```diff
@@ -51,9 +51,18 @@ sealed trait BaseKvRdd {
   val baseFlatSchema: StructType = StructType(keySchema ++ valueSchema)
   def flatSchema: StructType = if (withTime) StructType(baseFlatSchema :+ timeField) else baseFlatSchema
   def flatZSchema: api.StructType = flatSchema.toChrononSchema("Flat")
-  lazy val keyToBytes: Any => Array[Byte] = AvroConversions.encodeBytes(keyZSchema, GenericRowHandler.func)
+  // For global aggregations, use plain UTF-8 bytes for a dummy key (no need for an avro schema)
```
rgruener, line 54: I honestly think this is a bit of a hack but it is necessary since most KVStore implementations will require a non-empty key.

Global aggregations also pretty much require the user to cache the stored results since they will be fetched with every online request. It would be helpful to be more opinionated on how they are stored but I don't want to broadly affect the code base

PR #1076, opened 2025-11-25
Ensure topic existence for the JoinSource

pengyu-hou:
## Summary

Added validation to ensure that when a GroupBy uses a JoinSource, the parent join must have a topic defined for left source.

For the batch only chaining use cases, users can leverage the utils method like get output table and use a regular event source or entity source. 

## Why / Goal

Make the config consistent with Scala Join source runner: https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/streaming/JoinSourceRunner.scala#L204

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers

### DIFF
--- a/api/py/ai/chronon/group_by.py
+++ b/api/py/ai/chronon/group_by.py
@@ -240,12 +240,16 @@ def validate_group_by(group_by: ttypes.GroupBy):
                 assert not utils.is_streaming(src), "SNAPSHOT accuracy should not be specified for streaming sources"
         elif src.joinSource:
             # When using a join as a source, the parent join must have a topic for streaming
+            # If you want to chain batch only joins, use regular EventSource/EntitySource instead
             parent_join = src.joinSource.join
             parent_left = parent_join.left if parent_join else None
             assert parent_left and utils.is_streaming(parent_left), (
                 "When using a JoinSource as the source for a GroupBy, the parent join must have a topic. "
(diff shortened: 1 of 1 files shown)


api/py/ai/chronon/group_by.py
```diff
@@ -238,6 +238,15 @@ def validate_group_by(group_by: ttypes.GroupBy):
                 )
             else:
                 assert not utils.is_streaming(src), "SNAPSHOT accuracy should not be specified for streaming sources"
+        elif src.joinSource:
+            # When using a join as a source, the parent join must have a topic for streaming
```
hzding621, line 242: I don't think this is always true. We can have offline-only chaining too.

hzding621: After offline discussion with @pengyu-hou we think that we should reserve chaining for streaming use cases. For offline only, it's more simple to just use the table name of upstream join table.

hzding621: @pengyu-hou can we add more context to the error message and suggest user to use get_output_tables?

pengyu-hou: Done!

Shiyinghaha (approved):

PR #1078, opened 2025-11-25
[Chronon] Throwing exception when encountering duplicates in join part calculation

Shiyinghaha:
## Summary

User concern: chronon occasionally randomly produce duplicate records for joins, retry will solve the issue and hard to replicate, solution is to add this alert to fail the job if duplication occurs so that we will be notified and auto-retry will likely resolve the job

Throwing exception when encountering duplicates (should be distinct on `key col` or `key col + ts`) in join part calculation

Default is not enabled, users can choose enable by setting `joinPartUniquenessCheck` to true

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @yuli-han @pkundurthy @hzding621

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
+++ b/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
@@ -345,20 +345,22 @@ abstract class JoinBase(joinConf: api.Join,
       rightDfWithDerivations.prettyPrint()
     }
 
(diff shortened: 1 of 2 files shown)


pengyu-hou (approved):

hzding621 (approved):
Let's do an integration test pass before merging

PR #1118, opened 2026-05-08
Fix broadcast fetch failures when CatalystUtil initializes on executor JVMs in chained streaming GroupBys

yuli-han:
## Summary

**Problem**

When running a double-chained streaming GroupBy — where a GroupBy's source is a Join, and that Join's right-parts include GroupBys that themselves have non-trivial SQL derivations (e.g. CONCAT_WS, CASE WHEN, etc.) — the streaming job fails immediately with:

`Failed to get broadcast_1_piece0 of broadcast_1 across all executors.`

**Root Cause**

CatalystUtil.session is a lazy val that spins up a local[*] SparkSession, fine on the driver, but dangerous on executors. 

In the chained streaming path, GroupByServingInfoParsed.deriveFunc is a @transient lazy val that re-initializes after deserialization, and for GroupBys with non-trivial derivations, that re-initialization hits CatalystUtil.session.

When local[*] starts on an executor JVM, it creates a second SparkContext that registers itself as "driver" in the BlockManager routing table, overwriting the real driver's address and causing all broadcast fetches to fail.

**Fix**
Save SparkEnv.get (an AtomicReference, accessible from any thread) before creating the local session, then restore it immediately after. This undoes the SparkEnv overwrite caused by the new local[*] SparkContext before any tasks attempt to fetch broadcast blocks.

```
scalaval savedEnv = SparkEnv.get
val spark = SparkSession.builder().master("local[*]") ... .getOrCreate()
if (savedEnv != null && (SparkEnv.get ne savedEnv)) {
  SparkEnv.set(savedEnv)
}
```

The local session is still created (CatalystUtil needs it for Catalyst planning and codegen), but the executor's real SparkEnv is restored before any damage propagates. On unit test JVMs savedEnv is null, so the restore is a no-op — all existing tests continue to pass.

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

Integration tested
## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @Shiyinghaha

### DIFF
--- a/online/src/main/scala/ai/chronon/online/CatalystUtil.scala
+++ b/online/src/main/scala/ai/chronon/online/CatalystUtil.scala
@@ -31,6 +31,7 @@ import org.apache.spark.sql.execution.{
   RDDScanExec,
   WholeStageCodegenExec
 }
+import org.apache.spark.SparkEnv
 import org.apache.spark.sql.{SparkSession, types}
 
 import java.util.concurrent.{ArrayBlockingQueue, ConcurrentHashMap}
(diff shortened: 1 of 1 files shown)


pengyu-hou (approved):
Great find!

Shiyinghaha (approved):

PR #1119, opened 2026-05-11
[Chronon] Separate logging failure from blocking fetching, make chronon code able to handle struct for logging

Shiyinghaha:
## Summary

  - Isolate logging failures from the user-facing fetch path in `Fetcher.logResponse`. The existing `Try`-based guards only catch `NonFatal`, so fatal-typed throwables (e.g. `AssertionError` from `Row.to`) bypassed them and    
  failed the fetch. Wrap the body in an outer try/catch that re-throws true fatals (`VirtualMachineError`, `ThreadDeath`, `InterruptedException`, `LinkageError`, `ControlThrowable`) and swallows everything else with the        
  existing exception metric + error log.
  - Handle struct values that arrive as a `Map[String, Any]` in `Row.to`. Nested rows surfaced through Spark SQL transforms (e.g. `trip_reservation_data.reservation_details` in `ai_signals_platform/trip.user_history.v3`) reach 
  the avro encoder as `scala.collection.immutable.HashMap`, which previously hit `assert(extraneousRecord != null, ...)` because only `Array[Any]` and `util.ArrayList[Any]` were handled. Add `collection.Map[_, _]` and          
  `util.Map[_, _]` cases that emit values in struct-declared order, defaulting missing keys to `null`.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested

Tested against the user config in gateway machine with sampling = 100.0, fetching succeeded

```
SPARK_VERSION=3.2.1 python3 ~/.local/bin/run.py --mode=fetch -k '{"id_guest":xxxxx}' --conf-path production/joins/ai_signals_platform/trip.user_history.v3 -t join --chronon-jar ~/test/chronon-embedded.jar
```

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @yuli-han

### DIFF
--- a/online/src/main/scala/ai/chronon/online/Fetcher.scala
+++ b/online/src/main/scala/ai/chronon/online/Fetcher.scala
@@ -45,6 +45,7 @@ import scala.collection.mutable.ListBuffer
 import scala.collection.{Seq, mutable}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.ScalaJavaConversions.{IteratorOps, JListOps, JMapOps}
+import scala.util.control.ControlThrowable
 import scala.util.{Failure, Random, Success, Try}
 
 object Fetcher {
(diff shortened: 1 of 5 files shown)


online/src/main/scala/ai/chronon/online/Fetcher.scala
```diff
@@ -453,6 +454,24 @@ class Fetcher(val kvStore: KVStore,
...
+    // Wrapping logging with try catch to separate logging failure from the user-facing fetch response.
+    try {
+      logResponseInternal(resp)
+    } catch {
+      case e: VirtualMachineError  => throw e
+      case e: ThreadDeath          => throw e
+      case e: InterruptedException => throw e
+      case e: LinkageError         => throw e
+      case e: ControlThrowable     => throw e
+      case e: Throwable =>
```
pengyu-hou: can we simplify this section?

pengyu-hou (commented):
can you add unit test ？

online/src/main/scala/ai/chronon/online/Fetcher.scala
Shiyinghaha: updated

Shiyinghaha:
> can you add unit test ？

added unit test

pengyu-hou (approved):

yuli-han (approved):

Issue #1124, opened 2026-05-30
[Bug] TEMPORAL fetch can silently fall back to batch-only data when streaming KV reads fail


fallintoplace:
I was tracing through the online fetch path and found something that looks risky in TEMPORAL GroupBy fetches.

In `online/src/main/scala/ai/chronon/online/FetcherBase.scala`, the streaming response is built like this:

`streamingRequestOpt.map(responsesMap.getOrElse(_, Success(Seq.empty)).getOrElse(Seq.empty))`

Because of that, a per-request failure from KV on the streaming side gets turned into `Seq.empty`, which is the same thing the code would use for "there just isn't any streaming data".

So if the batch request succeeds but the streaming request comes back as `Failure(...)`, the request can still succeed and return batch-only values.

Why this seems problematic:
- it hides the difference between "no streaming rows" and "streaming read failed"
- TEMPORAL fetches can return plausible but stale values during a partial KV outage
- from the caller side, this looks like a normal success instead of a degraded response

A small repro would be:
1. Create a TEMPORAL GroupBy with both batch and streaming data.
2. Return `Success(...)` for the batch `GetResponse`.
3. Return `Failure(new RuntimeException("kv down"))` for the streaming `GetResponse`.
4. `fetchGroupBys` currently treats the streaming side like an empty sequence and proceeds.

Expected behavior:
- either propagate the streaming failure for TEMPORAL fetches
- or surface an explicit degraded/error signal instead of silently treating it as empty streaming input

I might be missing some intended semantics here, but if `Failure` is meant to carry real KV errors, this seems worth tightening up.

Happy to put together a focused test as well if that would help.

PR #1125, opened 2026-05-30
Fail temporal fetch when streaming KV reads fail

fallintoplace:
Addresses #1124.

## Summary
- fail TEMPORAL GroupBy fetches when the streaming KV request fails or is missing instead of treating it like empty streaming data
- add a regression test for the mixed case where the batch KV read succeeds but the streaming KV read fails

## Why
`FetcherBase` currently collapses a failed streaming KV response into `Seq.empty`. That makes a real streaming read failure look the same as "there just wasn't any streaming data", which can let a TEMPORAL fetch succeed with batch-only values.

This change takes the stricter path: if the streaming side fails, the TEMPORAL fetch fails too.

## Validation
- attempted `/tmp/bazelisk test //spark:test --test_filter=ai.chronon.spark.test.FetcherTest#testTemporalGroupByStreamingFailurePropagates`
- Bazel analysis was blocked by an external dependency fetch failure for `org.pentaho:pentaho-aggdesigner-algorithm:5.1.5-jhyde`

### DIFF
--- a/online/src/main/scala/ai/chronon/online/FetcherBase.scala
+++ b/online/src/main/scala/ai/chronon/online/FetcherBase.scala
@@ -496,8 +496,15 @@ class FetcherBase(kvStore: KVStore,
                   case Some(cachedResponse: CachedBatchResponse) => cachedResponse
                 }
 
-              val streamingResponsesOpt =
-                streamingRequestOpt.map(responsesMap.getOrElse(_, Success(Seq.empty)).getOrElse(Seq.empty))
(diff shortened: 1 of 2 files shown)


gsarthakdev:
Hey @fallintoplace, I came across your PR & issue #1124, it looks like some cci tests have failed which is likely why the PR hasn't been merged yet...? I'm also curious if the maintainers don't see this as a valid issue, in that case we can close it to keep the GitHub clean

 cc: @pengyu-hou, @nikhil-zlai

PR #1128, opened 2026-06-11
Add incremental SNAPSHOT GroupBy aggregation

pengyu-hou:
## Summary

Adds an opt-in `is_incremental` mode for **SNAPSHOT-accuracy event** GroupBy backfills. Instead of re-scanning the full window of raw events on every backfill step, per-day partial aggregates (IRs) are computed once and cached in a `<table>_daily_inc` table, then reused. This turns the dominant cost (raw-event hop aggregation) into a write-once read-many step.

Validated end-to-end on a ~500M-rows/day, 730-day-window source: a warm daily run completes in 
~30 min vs ~6 h for the equivalent full backfill (**~10× wall-clock**), producing output that matches the non-incremental path on all deterministic columns .

  - `is_incremental=True` on the GroupBy sets a new thrift field `isIncremental`.
  - **Build** (`computeIncrementalDf`): fills the `_daily_inc` IR table for `[ds − maxWindow, ds]`, idempotent via
  `unfilledRanges` (no-op when already complete).
  - **Read** (`fromIncrementalDf`): returns a GroupBy whose `hopsAggregate` is served from the cached IRs instead of raw events; `snapshotEvents` then windows them as usual.
  - Build and read are split so the daily table can later be reused by upload and join (see follow-ups). `fromIncrementalDf` is a mode-agnostic adapter; `buildIfMissing` controls ensure-then-read (ad hoc) vs read-only (production behind a producer node).

image

## Follow-ups (separate PRs)

  - Snapshot-events incremental **upload** (producer node + ensure-then-read).
  - **Join** snapshot right-parts consuming the same `_daily_inc` table.

## Why / Goal

Even though Chronon can scale sublinearly, the limitation is usually within the same Spark session. 

This change will enable persisting the intermediate result so that for the large backfills, it does not have to read the raw events.

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested
- [x] tested on gateway machine and compare with standard output 
## Checklist
- [ ] Documentation update

## Reviewers
@yuli-han @Shiyinghaha @kambstreat

### DIFF
--- a/api/py/ai/chronon/group_by.py
+++ b/api/py/ai/chronon/group_by.py
@@ -390,6 +390,7 @@ def GroupBy(
     historical_backfill: Optional[bool] = None,
     deprecation_date: Optional[str] = None,
     description: Optional[str] = None,
+    is_incremental: Optional[bool] = None,
     **kwargs,
 ) -> ttypes.GroupBy:
     """
@@ -608,6 +609,7 @@ def GroupBy(
         backfillStartDate=backfill_start_date,
         accuracy=accuracy,
         derivations=derivations,
+        isIncremental=is_incremental,
     )
     validate_group_by(group_by)
     return group_by
--- a/api/py/ai/chronon/group_by.py
+++ b/api/py/ai/chronon/group_by.py
@@ -255,6 +255,16 @@ def validate_group_by(group_by: ttypes.GroupBy):
             if contains_windowed_aggregation(aggregations):
                 assert query.timeColumn, "Please specify timeColumn for entity source with windowed aggregations"
 
+    # is_incremental currently only supports SNAPSHOT accuracy with event sources.
+    # (Temporal accuracy and entity sources may be supported later.)
+    if group_by.isIncremental:
+        assert group_by.accuracy == Accuracy.SNAPSHOT, (
+            "is_incremental is only supported for SNAPSHOT accuracy group bys"
+        )
+        assert all([s.events for s in sources]), (
+            "is_incremental is only supported for event sources"
+        )
+
     column_set = None
     # all sources should select the same columns
     for i, source in enumerate(sources[1:]):
--- a/api/py/ai/chronon/group_by.py
+++ b/api/py/ai/chronon/group_by.py
@@ -258,12 +258,15 @@ def validate_group_by(group_by: ttypes.GroupBy):
     # is_incremental currently only supports SNAPSHOT accuracy with event sources.
     # (Temporal accuracy and entity sources may be supported later.)
     if group_by.isIncremental:
-        assert group_by.accuracy == Accuracy.SNAPSHOT, (
-            "is_incremental is only supported for SNAPSHOT accuracy group bys"
-        )
         assert all([s.events for s in sources]), (
             "is_incremental is only supported for event sources"
         )
+        # accuracy is optional: when unset it is inferred as SNAPSHOT unless a streaming
+        # source is present (which infers TEMPORAL). Accept the inferred-SNAPSHOT case.
+        is_snapshot = group_by.accuracy == Accuracy.SNAPSHOT or (
+            group_by.accuracy is None and not any([utils.is_streaming(s) for s in sources])
+        )
+        assert is_snapshot, "is_incremental is only supported for SNAPSHOT accuracy group bys"
 
     column_set = None
     # all sources should select the same columns
--- a/api/src/main/scala/ai/chronon/api/Extensions.scala
+++ b/api/src/main/scala/ai/chronon/api/Extensions.scala
@@ -97,7 +97,7 @@ object Extensions {
     def cleanName: String = metaData.name.sanitize
 
     def outputTable = s"${metaData.outputNamespace}.${metaData.cleanName}"
-
+    def incrementalOutputTable = s"${metaData.outputNamespace}.${metaData.cleanName}_daily_inc"
     def preModelTransformsTable = s"${metaData.outputNamespace}.${metaData.cleanName}_pre_mt"
     def outputLabelTable = s"${metaData.outputNamespace}.${metaData.cleanName}_labels"
     def outputFinalView = s"${metaData.outputNamespace}.${metaData.cleanName}_labeled"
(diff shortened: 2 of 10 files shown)


api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -178,6 +178,10 @@ object Extensions {
 
     def outputColumnName =
       s"${aggregationPart.inputColumn}_$opSuffix${aggregationPart.window.suffix}${bucketSuffix}"
+
+    def incrementalOutputColumnName =
+      s"${aggregationPart.inputColumn}_$opSuffix${bucketSuffix}"
```
yuli-han, line 183: will this bring duplicate columns if the groupBy has aggregated features over different windows?

spark/src/main/scala/ai/chronon/spark/GroupBy.scala
```diff
@@ -688,15 +722,272 @@ object GroupBy {
...
+  def computeIncrementalDf(
+      groupByConf: api.GroupBy,
+      range: PartitionRange,
+      tableUtils: TableUtils,
+      incrementalOutputTable: String,
+      stepDays: Option[Int] = None
+  ): (PartitionRange, Seq[api.AggregationPart]) = {
+
+    val tableProps: Map[String, String] = Option(groupByConf.metaData.tableProperties)
+      .map(_.toScala)
```
yuli-han, line 800: can we add the chronon table tag and create a new table type in https://github.com/airbnb/chronon/blob/main/api/src/main/scala/ai/chronon/api/Constants.scala#L76? 
```
Map(                                                                                                                                                                                
      Constants.ChrononGenerated -> "true",                                                                                                                                                                      
      Constants.ChrononTableType -> Constants.TableType.GroupBy  // ✓ tagged                                                                                                                                     
    )   
```      
Example: https://github.com/airbnb/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/GroupBy.scala#L712

api/py/ai/chronon/group_by.py
```diff
@@ -255,6 +255,16 @@ def validate_group_by(group_by: ttypes.GroupBy):
             if contains_windowed_aggregation(aggregations):
                 assert query.timeColumn, "Please specify timeColumn for entity source with windowed aggregations"
 
+    # is_incremental currently only supports SNAPSHOT accuracy with event sources.
+    # (Temporal accuracy and entity sources may be supported later.)
+    if group_by.isIncremental:
+        assert group_by.accuracy == Accuracy.SNAPSHOT, (
```
yuli-han: can you check whether this accuracy is set by users or can be inferred? Does it work if the user doesn't set the accuracy explicitly?

spark/src/main/scala/ai/chronon/spark/GroupBy.scala
```diff
@@ -688,15 +722,272 @@ object GroupBy {
...
+    new GroupBy(
+      groupByConf.getAggregations.toScala,
+      groupByConf.getKeyColumns.toScala,
+      sourceDf,
+      () => null
+    ) {
+      // Serve the precomputed daily IRs instead of aggregating raw events.
+      override def hopsAggregate(minQueryTs: Long,
+                                 resolution: Resolution): RDD[(KeyWithHash, HopsAggregator.OutputArrayType)] =
+        incrementalHops
```
yuli-han, line 972: should we verify resolution == DailyResolution here?

api/src/main/scala/ai/chronon/api/Extensions.scala
pengyu-hou: incrementalOutputColumnName intentionally drops the window suffix so `SUM(price, [30d,90d,365d,730d])` → one `price_sum` column, not four.

api/py/ai/chronon/group_by.py
pengyu-hou: Good catch!

spark/src/main/scala/ai/chronon/spark/GroupBy.scala
pengyu-hou: added!

spark/src/main/scala/ai/chronon/spark/GroupBy.scala
pengyu-hou: Done!

api/src/main/scala/ai/chronon/api/Extensions.scala
pengyu-hou: also added dedup logic here.

spark/src/main/scala/ai/chronon/spark/GroupBy.scala
```diff
@@ -688,15 +749,272 @@ object GroupBy {
...
+
+    val incrementalOutputTable = groupByConf.metaData.incrementalOutputTable
+    assert(incrementalOutputTable != null,
+           s"incrementalOutputTable is not set for GroupBy: ${groupByConf.metaData.name}")
+
+    val (incrementalQueryableRange, aggregationParts) =
+      if (buildIfMissing) {
+        // ensure-then-read: fill any holes (no-op if already complete), then read.
+        computeIncrementalDf(groupByConf, range, tableUtils, incrementalOutputTable, stepDays)
+      } else {
```
Shiyinghaha, line 933: What would be the expected behavior if `buildIfMissing` is set to false? Is it possible for unfilled range to exist in this scenario and if yes, how should we handle it?

pengyu-hou: This flag is used for differentiating the following two mode:
-  standalone backfill/upload mode
- in production, we have a GB incremental producer, the two backfill and upload consumer could read from its output to avoid racing condition. 
The diagram is like:

Image

spark/src/main/scala/ai/chronon/spark/GroupBy.scala
yuli-han: Can we added it inside the function hopsAggregate? This is to avoid calling this for groupBy upload for TEMPORAL and passes FiveMinuteResolution

yuli-han (approved):

Shiyinghaha (approved):

PR #1129, opened 2026-06-12
Enable snapshot-events GroupBy upload to read from the incremental table

pengyu-hou:
## Summary

Serve the **GroupBy upload** from the cached daily IR table (`<table>_daily_inc`) instead of re-scanning raw events, reusing the incremental machinery already used by backfill (#1128). Applies to **SNAPSHOT accuracy + event sources**; all other accuracy/data-model paths are unchanged.

## What changed

- **Upload reads the IR cache** — `GroupByUpload.run`: for `SNAPSHOT + events + is_incremental`, build the upload GroupBy via `fromIncrementalDf` (ensure-then-read) rather than scanning raw events. A `require()` rejects incremental on any non-snapshot / non-events combination that reaches upload.
- **Cold-build chunking** — single source of truth `GroupBy.DefaultStepDays` (= 30), shared by the CLI `--step-days` default and the incremental build, so a first/cold build (e.g. the single-day upload path) always chunks and never does a monolithic window-wide write. Hole detection is source-aware (skips days with no source data).
- **Orchestration** — backfill and upload are both ensure-then-read consumers; in production they run daily, ordered `backfill → upload`, so they never write the IR table concurrently. 

## Testing

Deterministic data-quality tests:
- daily IR build covers every event day, including range boundaries
- incremental snapshot output matches the non-incremental path at the window tail boundary
- incremental upload output matches a non-incremental upload (KV rows)
- `testIncrementalFirstLast` uses unique values so `TOP_K`/`BOTTOM_K` are tie-free (deterministic)

`GroupByIncrementalTest` + `GroupByUploadTest` pass on Scala 2.12 and 2.13.

## Reviewer
@yuli-han @Shiyinghaha

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Driver.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Driver.scala
@@ -678,7 +678,7 @@ object Driver {
     }
 
     def run(args: Args): Unit = {
-      GroupByUpload.run(parseConf[api.GroupBy](args.confPath()), args.endDate())
+      GroupByUpload.run(parseConf[api.GroupBy](args.confPath()), args.endDate(), stepDays = args.stepDays.toOption)
     }
   }
 
--- a/spark/src/main/scala/ai/chronon/spark/GroupBy.scala
+++ b/spark/src/main/scala/ai/chronon/spark/GroupBy.scala
@@ -786,7 +786,8 @@ object GroupBy {
     * range a reader must scan and the unwindowed aggregation parts. Distinct from the instance-level
     * `computeIncrementalDf`, which is the per-hole worker that writes a single sub-range.
     *
-    * @param stepDays optional chunking of each hole into stepped sub-ranges (restartable, bounded memory)
+    * @param stepDays chunk size for filling holes. Supplied by the CLI --step-days (default 30, see
+    *                 OfflineSubcommand in Driver.scala); None (caller omits it) fills in one write.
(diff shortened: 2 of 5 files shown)


docs/source/Incremental_GroupBy.md
```diff
@@ -0,0 +1,261 @@
...
+  partition D+1 is only captured once D+1 exists. For isolated single-day hole-fills
+  of historical days this can under-count; a full build reads every partition and is
+  unaffected. A bounded forward-read (e.g. +1 day) is a possible future enhancement.
+
+## 8. Future work
+
+- **Join incremental right-parts:** snapshot right-parts consume `fromIncrementalDf`;
+  a GroupBy used by its own backfill plus N joins builds `_daily_inc` once and all
+  consumers read it.
+- **Weekly rollup tier:** for very long windows, a coarse weekly hop band could cut
```
abbywh: We do 30d, and 180d, though technically you can configure your own. I find this saturates the hops quite well.

pengyu-hou: agreed. I was thinking of more like weekly and monthly to speed up. but I want to handle it later once the daily ones are stable. The larger the window is, the less I/O cost 

Thanks @abbywh for taking a look

yuli-han (approved):

Shiyinghaha (approved):

PR #1142, opened 2026-08-18
Adding the option to skip intermediate table materialization 

Shiyinghaha:
## Summary

Adding the option to skip intermediate table materialization. When `materializeJoinParts` is set to `false` and no join part parallelization is enabled, no join part intermediate table will be materialized. If `materializeJoinParts` is set to `false` and join part parallelization is enabled, it will automatically fall back to materializing join parts. 

## Why / Goal

Improve complex join computation efficiency. 

## Test Plan

- [x] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested (See internal PRs for details), est. runtime saving is more than 50% for complex joins. 

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @yuli-han

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
+++ b/spark/src/main/scala/ai/chronon/spark/JoinBase.scala
@@ -45,13 +45,19 @@ abstract class JoinBase(joinConf: api.Join,
                         unsetSemanticHash: Boolean) {
   @transient lazy val logger = LoggerFactory.getLogger(getClass)
   assert(Option(joinConf.metaData.outputNamespace).nonEmpty, s"output namespace could not be empty or null")
-  require(
-    tableUtils.materializeJoinParts || selectedJoinParts.isEmpty,
-    "spark.chronon.join.part.materialize=false is not compatible with the parallelized join flow. " +
-      "That flow hands joinPart results between separate Spark jobs through the join part tables, so it needs " +
-      "them materialized. Either drop --selected-join-parts and run the monolithic backfill, or leave " +
-      "spark.chronon.join.part.materialize at its default of true."
-  )
+
+  // The parallelized join flow hands joinPart results between separate Spark jobs through the join part tables,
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/ai/chronon/spark/JoinBase.scala
```diff
@@ -45,6 +45,19 @@ abstract class JoinBase(joinConf: api.Join,
...
+  // so those tables have to exist for it to work at all. When that flow is in use we materialize regardless of
+  // the flag instead of failing the job: the flag is typically set once for a cluster, while the flow is chosen
+  // per run, so a job that cannot honor the flag should still produce correct output.
+  private val parallelizedJoinPartFlow: Boolean = selectedJoinParts.isDefined
+  val materializeJoinParts: Boolean = tableUtils.materializeJoinParts || parallelizedJoinPartFlow
+  if (parallelizedJoinPartFlow && !tableUtils.materializeJoinParts) {
+    logger.warn(
+      "spark.chronon.join.part.materialize=false is not supported by the parallelized join flow, which hands " +
+        "joinPart results between separate Spark jobs through the join part tables. Falling back to writing " +
+        "them. Drop --selected-join-parts to run the monolithic backfill with the flag honored.")
```
yuli-han, line 59: Should this be an error or a warn? We cannot run parallelized join flow with no materialized join parts right?

Shiyinghaha: we cannot but the code will fall back automatically to materialize join parts

but can be changed to logger.error

Shiyinghaha: meanwhile airchat says: The job is fine. error conventionally means output is compromised or something failed. Here the code deliberately recovers and produces correct results. Nothing downstream is wrong and nobody needs to act.

so maybe its ok to keep logger.warn

yuli-han (approved):

pengyu-hou (approved):

PR #1146, opened 2026-08-28
[Streaming] Limit checkpoint to the Spark application

larissacopel:
## Summary

Currently, Chronon's streaming checkpoint is built from `spark.chronon.stream.checkpoint_base_dir` + the GroupBy's name. This path is static -- it depends only on the GroupBy config, not on the specific Spark run. Every retry of the same streaming job (whether it's in the same driver retrying, or a brand-new job submission after a restart) resolver to the exact same checkpoint directory.

**New behavior:** the checkpoint path now has the Spark application ID appended as a suffix, making it unique per Spark application rather than shared across every run of the job. This will guarantee that new tasks from within the same application don't have data gaps.

## Why / Goal

The current behavior doesn't cover data gaps between new Spark applications and this will match the current behavior while allowing us to benefit from the checkpointing for the data gaps between Spark tasks from within the same application.

## Test Plan

- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested

Executed a job manually with this change and worked as expected.

## Checklist
- [ ] Documentation update

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/streaming/CheckpointConfig.scala
+++ b/spark/src/main/scala/ai/chronon/spark/streaming/CheckpointConfig.scala
@@ -27,6 +27,6 @@ object CheckpointConfig {
   // to avoid changing behavior for callers that haven't opted in.
   def location(groupByConf: api.GroupBy, session: SparkSession): Option[String] =
     session.conf.getOption("spark.chronon.stream.checkpoint_base_dir").map { baseDir =>
-      s"$baseDir/${groupByConf.metaData.nameToFilePath}/"
+      s"$baseDir/${groupByConf.metaData.nameToFilePath}/${session.sparkContext.applicationId}/"
     }
 }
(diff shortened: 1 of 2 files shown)


pengyu-hou (approved):

yuli-han (approved):

nikhil-zlai:
if we restart the streaming job, the second app wont pick up where the first job stopped - due to this change, right?

larissacopel:
> if we restart the streaming job, the second app wont pick up where the first job stopped - due to this change, right?

Yes, that's correct. That's the current behavior and I'm only matching it to allow us to use the checkpoint only from within the same application.

nikhil-zlai:
i am pretty sure that is problematic for prod jobs. @pengyu-hou can clarify perhaps.

pengyu-hou:
> i am pretty sure that is problematic for prod jobs. @pengyu-hou can clarify perhaps.

Hi @nikhil-zlai, thanks for reviewing the change. 

This is actually indeed our current prod behavior. We switched to read from the latest offset after restarting when we keep running into issues with the KV store due to the write spike. And it stayed this way until now. 

There could be improvements towards this: write smooth after job initialization to mitigate such issues. But the current approach is also easier to operate. Let me know what you think. Thanks again!

nikhil-zlai:
gotcha! so we tolerate write holes during restarts within the day? 

reasonable tradeoff, perhaps the ratings use case wont tolerate it, but most ml use-cases maybe fine.

pengyu-hou:
> gotcha! so we tolerate write holes during restarts within the day?
> 
> reasonable tradeoff, perhaps the ratings use case wont tolerate it, but most ml use-cases maybe fine.

@nikhil-zlai oh yeah, the trade-off was actually made by you and Varant before the rating use case back in 2021 or 2022.

PR #1151, opened 2026-09-09
[Streaming] Bound chained fetch fan-out and make inference retries configurable

pengyu-hou:
[AIRCHAT] Gives chained streaming jobs a fixed rows-per-fetch and a per-executor in-flight bound, and makes model inference retries a per-model policy with backoff.

## Summary

Two related changes to the model-transform path of chained streaming jobs.

**Spark streaming (`JoinSourceRunner`, new `ChunkedFetch`)**
- `spark.chronon.stream.chain.fetch_chunk_size`: rows per chained fetch for both the base-join and the model-transform step. Default 0 keeps one fetch per partition.
- `spark.chronon.stream.chain.max_inflight_per_executor`: outstanding chunk fetches per executor JVM, shared across all tasks via a semaphore in `LocalIOCache`. Default 0 leaves it unbounded. Waiting for a permit counts against `timeout_millis`, so a saturated executor times out the same way a slow backend would.
- The model-transform step now emits `chain.request_batch_size` too, so partition size at that step is observable directly.

**Online (`FetcherModelUtils`)**
- The fixed `withRetry(maxRetries = 2)` with no delay becomes a `RetryPolicy` read from `model.inferenceSpec.modelBackendParams`: `chronon_inference_max_retries`, `chronon_inference_retry_backoff_ms` (exponential, 20% jitter, capped by `chronon_inference_retry_max_backoff_ms`), and `chronon_inference_non_retryable_exceptions` (comma-separated substrings matched against class names and messages through the cause chain). Defaults are the historical two immediate retries.
- `chronon_inference_chunk_size` optionally splits the merged per-model request so a failure nulls the outputs of one chunk instead of every row in the partition.
- New `inference.input_count` distribution per backend call.

## Why / Goal

On 2026-09-08 two chained streaming jobs at Airbnb fell behind while their Manifold model backend was slow. Partitions grew from ~20 rows to ~325, the backend split each partition into ~10 parallel 32-row calls, and with 144 concurrent tasks that meant ~1,400 in-flight calls plus up to six attempts per call between Chronon's retry and the client's own retry. Model-level exceptions sat at 16-37% for seven hours. The loop closes because lag grows partition size, partition size grows backend concurrency, and backend latency grows lag. Fixing rows per fetch and bounding in-flight fetches removes the coupling; backoff and a non-retryable list stop retries from adding load to a backend that has just asked for less.

Companion changes: treehouse `pengyu--manifold-bounded-concurrency` (adds `max_concurrent_batches` to the Manifold backend) and ml_models `pengyu/gr-history-state-chain-bounding` (turns the settings on for the two affected jobs).

## Test Plan
- [x] Added Unit Tests
  - `FetcherModelUtilsRetryTest` (9): defaults, param parsing, non-retryable matching through the cause chain, backoff bounds, retry count, success after retry, non-retryable short-circuit, delayed retry timing, chunk size parsing.
  - `ChunkedFetchTest` (5): whole-partition path, chunk order and sizes, semaphore bound on concurrency, permit release on failure, permit wait counting against the timeout.
- [x] Covered by existing CI
  - `ModelTransformsTest` and `ChainingFetcherTest` still pass.
- [ ] Integration tested
  - Will validate on the two GR history-state jobs once a runtime with this change is pinned.

## Checklist
- [ ] Documentation update


spark/src/main/scala/ai/chronon/spark/streaming/JoinSourceRunner.scala
```diff
@@ -434,6 +458,17 @@ class JoinSourceRunner(groupByConf: api.GroupBy, conf: Map[String, String] = Map
     }
   }
 
+  @transient private lazy implicit val chunkExecutionContext: scala.concurrent.ExecutionContext =
```
yuli-han, line 461: `chunkExecutionContext` here is declared as a per-instance `@transient lazy val` rather than going through `LocalIOCache` like `fetcher`/`kvStore`/`inflightFetches` above. Per the class comment ("micro batching destroys and re-creates these objects repeatedly through ForeachBatchWriter and MapFunction"), that means `FlexibleExecutionContext.buildExecutionContext` (a dedicated ThreadPoolExecutor: corePoolSize=20, maxPoolSize=1000, 600s keep-alive, core threads never time out) gets rebuilt on every micro-batch/task recreation of JoinSourceRunner, with no shutdown path.

It's also not gated behind fetch_chunk_size > 0: both call sites of fetchChunked() (base-join fetch and model-transform fetch) invoke it unconditionally, and since it's an implicit lazy val supplied at the call site, it gets forced/evaluated regardless of which branch ChunkedFetch.run takes internally -- including the disabled-by-default (fetch_chunk_size = 0) case.

Given this PR's goal is to bound resource usage under load, this reintroduces the same class of problem LocalIOCache exists to prevent. Suggest hoisting a shared ExecutionContext into LocalIOCache (same getOrSet pattern as getOrSetInflightFetches) instead of a per-instance lazy val.

PR #1156, opened 2026-09-17
[Online] Bind CatalystUtil evaluators to the Chronon session conf on every thread

pengyu-hou:
## Summary
`CatalystUtil` configures `spark.sql.legacy.timeParserPolicy=LEGACY` and `spark.sql.session.timeZone=UTC` on its own local `SparkSession`, but Catalyst reads `SQLConf.get` during codegen and expression evaluation. On a non-task thread that resolves to the thread's *active* `SparkSession`, or, if none is set, to Spark's default fallback conf (`EXCEPTION` policy).

`PoolMap` seeds two evaluators per derivation set and constructs additional ones on the calling request thread whenever the pool is empty. A request thread has no active session, so those evaluators were code-generated with the default policy: `UnixTime.doGenCode` forces the lazy `formatterOption`, and `TimestampFormatter.getFormatter` bakes an `Iso8601TimestampFormatter` into the generated class. `to_unix_timestamp` on values like `2026-09-11 02:44:02.560` then throws `SparkUpgradeException`, and because the evaluator is returned to the pool, the failures persist intermittently until the next redeploy.

This PR adds `CatalystUtil.withSession`, which sets the Chronon session as the current thread's active session, runs the block, and restores the previous active session. It wraps evaluator construction (`initialize`) and the per-row transform calls.

Design notes:
- No-op on executors (`TaskContext` set): `SQLConf.get` uses task-local properties there.
- Deliberately does not use the JVM-global `SQLConf.setSQLConfGetter`. `JoinSourceRunner` hosts `CatalystUtil` inside a live Spark driver, and a global getter would redirect that job's own planning to the local session conf. Restoring the previous active session keeps the driver's session intact.
- Compiles unchanged on all cross builds (Scala 2.11 / Spark 2.4, 2.12 / 3.1, 2.13 / 3.2).

## Why / Goal
Prism (payments online serving, Spark 3.1.1) saw intermittent `SparkUpgradeException` followed by `derivation_rename_exception` on join `payments.homes_payin_chargeback.v1.online` once traffic increased. Symptoms matched exactly: healthy for 1-2 days after deploy, then sticky intermittent failures where retries succeed; Prophet on Spark 2.4 unaffected. Consumers should not need a service-side workaround such as installing a global SQLConf getter before creating the fetcher.

## Test Plan
- [x] Added Unit Tests: `testTimestampParsingIsConfSafeOnThreadWithoutActiveSession` builds a `CatalystUtil` on a thread with no active session and parses a millisecond-suffixed timestamp. Before this change it fails with the production `SparkUpgradeException`; after, it passes.
- [x] Covered by existing CI: `CatalystUtilTest` passes (35/35) on Scala 2.12 / Spark 3.1.1 and Scala 2.13 / Spark 3.2.1; `online/compile` passes on Scala 2.11 / Spark 2.4; `spark_uber/compile` passes; scalafmt clean.
- [ ] Integration tested: pending a Prism deploy on a build including this change.

## Checklist
- [ ] Documentation update (not needed; internal behavior fix)


gsarthakdev:
I've seen this term being used frequently, what does "[AIRCHAT]" at the beginning of PRs mean? I tried looking through CONTRIBUTING.md & online but couldn't find a concrete answer.

pengyu-hou:
> I've seen this term being used frequently, what does "[AIRCHAT]" at the beginning of PRs mean? I tried looking through CONTRIBUTING.md & online but couldn't find a concrete answer.

Hey, it is our internal agentic coding tool.

gsarthakdev:
> > I've seen this term being used frequently, what does "[AIRCHAT]" at the beginning of PRs mean? I tried looking through CONTRIBUTING.md & online but couldn't find a concrete answer.
> 
> 
> 
> Hey, it is our internal agentic coding tool. 

Cool! I'll try seeing if there are any blogs/videos covering this, Airchat sounds interesting!

-Sarthak

EDIT: Found info online, cool pic does this look right:
image

yuli-han (approved):

Shiyinghaha (approved):

PR #1157, opened 2026-09-18
Parallelize MetadataExporter's per-entity analysis

yuli-han:
## Summary

MetadataExporter.processEntities and enrichEmbeddedGroupBys process every GroupBy/Join/StagingQuery sequentially, and each entity's analysis is dominated by driver-bound metastore/catalog round-trip latency rather than executor compute. On configs with hundreds of embedded GroupBys (e.g. large Joins), this makes metadata-export jobs take multiple hours despite executors sitting mostly idle.

Add a parMap helper, mirroring the existing tableUtils.joinPartParallelism thread-pool pattern used for join-part backfills, and wire it into the three sequential .map loops. Controlled by a new
spark.chronon.metadata_export.parallelism conf, defaulting to 1 (today's exact sequential behavior), so this is opt-in and does not change behavior for existing jobs.

Also fixes a temp-view naming race in Extensions.range() (math.random -> UUID) that concurrent analyzeGroupBy calls would otherwise be able to hit.

## Test Plan

- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested

## Checklist
- [ ] Documentation update

## Reviewers
@pengyu-hou @Shiyinghaha


spark/src/main/scala/ai/chronon/spark/catalog/TableUtils.scala
```diff
@@ -402,6 +402,11 @@ case class TableUtils(sparkSession: SparkSession) {
   val materializeJoinParts: Boolean =
     sparkSession.conf.get("spark.chronon.join.part.materialize", "true").toBoolean
   val joinPartParallelism: Int = sparkSession.conf.get("spark.chronon.join.part.parallelism", "1").toInt
+  // number of entities (GroupBy/Join/StagingQuery configs) MetadataExporter analyzes concurrently.
+  // Each entity's analysis is a handful of small, driver-bound Spark jobs dominated by metastore/
+  // catalog round-trip latency rather than executor compute, so concurrency (not more executors)
+  // is what shortens wall-clock time here. Defaults to 1 (today's sequential behavior).
+  val metadataExportParallelism: Int = sparkSession.conf.get("spark.chronon.metadata_export.parallelism", "1").toInt
```
pengyu-hou, line 409: Is it safe to set the default to 10 or 20

pengyu-hou (approved):

spark/src/main/scala/ai/chronon/spark/catalog/TableUtils.scala
yuli-han: I keep the default at 1 on purpose, it makes this a no-op for every existing job unless a team explicitly opts in by setting spark.chronon.metadata_export.parallelism higher. Bumping the default to 10-20 would mean every team's metadata-export job starts hitting the shared Hive/Iceberg metastore concurrently with zero opt-in, and TableUtils' FormatProvider is pluggable, so I can't verify every team's custom FormatProvider is thread-safe under concurrency. Lowest-risk path is default=1, and anyone who wants the speedup can turn it on explicitly.

Shiyinghaha (approved):

